PR #27, opened 2017-11-17
adds ``exclude_types`` option, to exclude types from checking

richardasaurus:
Problem:

When writing testing code I often do something like this

```python
from mock import ANY

my_func(a=1, b=ANY)
```

Which makes sense when testing, but the type checked function will TypeError.  I tried to go through all my various `Union`'d types etc adding `ANY` and `Mock`, but it made a mess of the code. 

So I think this solution is cleaner overall. Let me know your thoughts.


richardasaurus:
@agronholm CI is failing as the package https://pypi.python.org/pypi/backports.typing has been removed from PYPI

agronholm:
Removed? That's almost unheard of. Ah well, looks like I need to update my code accordingly.

agronholm:
Alright, backports.typing support is gone now.

richardasaurus:
@agronholm CI passing now. is this change something you'd consider adding or?

agronholm:
The feature seems alright, but I'm -1 on this implementation. Although I generally don't like globals, I think that having a global registry of excluded types is the only reasonable approach here. By default it should probably contain `unittest.mock.Mock` at least.

richardasaurus:
can you provide an example of how you want it to look? then i can implement it that way

agronholm:
You could add a function like `register_excluded_type(cls)` which also makes sure that it's really a type. Then, the type checks will return when encountering an object of that type.

PR #39, opened 2018-07-30
Callmemo context

### LINKED ISSUE #38: _CallMemo is not using the scope in which the function call is declared
The following snippet helps explaining my issue:

```python

def tst():
    class A:

        @typechecked
        def clone(self) -> 'A':
            return A()


    val = A().clone()
    assert isinstance(val, A)

tst()
```

This makes sure that the class A is not available at module level because it is defined inside the function. This will result in the following stack trace:

```
Connected to pydev debugger (build 182.3684.100)
  File "/Applications/PyCharm.app/Contents/helpers/pydev/pydevd.py", line 1664, in <module>
    main()
  File "/Applications/PyCharm.app/Contents/helpers/pydev/pydevd.py", line 1658, in main
    globals = debugger.run(setup['file'], None, None, is_module)
  File "/Applications/PyCharm.app/Contents/helpers/pydev/pydevd.py", line 1068, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/Applications/PyCharm.app/Contents/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/Users/sjuul/workspace/typeguard/bla.py", line 14, in <module>
    tst()
  File "/Users/sjuul/workspace/typeguard/bla.py", line 11, in tst
    val = A().clone()
  File "/Users/sjuul/workspace/typeguard/typeguard.py", line 489, in wrapper
    memo = _CallMemo(func, args=args, kwargs=kwargs)
  File "/Users/sjuul/workspace/typeguard/typeguard.py", line 47, in __init__
    hints = get_type_hints(func)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/typing.py", line 1001, in get_type_hints
    value = _eval_type(value, globalns, localns)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/typing.py", line 260, in _eval_type
    return t._evaluate(globalns, localns)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/typing.py", line 464, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
NameError: name 'A' is not defined
```

The function `hints = get_type_hints(func)` cannot find the forward reference  `'A'` this is because typeguard using the current context and should use the context in which the function is defined in. This fixes it (I'll add a PR after)

```python
frame = inspect.stack()[2].frame
hints = get_type_hints(func, localns=frame.f_locals, globalns=frame.f_globals)
```

obeleh:
fixes #38


typeguard.py
```diff
@@ -42,7 +42,8 @@ def __init__(self, func: Callable, frame=None, args: tuple = None,
 
         self.type_hints = _type_hints_map.get(func)
         if self.type_hints is None:
-            hints = get_type_hints(func)
+            frame = inspect.stack()[2][0]
```
agronholm: Why are you ignoring the `frame` object passed to the constructor? This code should **not** rely on the current stack!

obeleh: You're right in most cases. This case though you cannot use the frame that is being extracted from "decoration time" the `class` doesn't exist in that frame yet. My implementation is incomplete. Let me commit the "normal case" first

obeleh:
Note that the test doesn't pass now since I'm not relying on the correct stack frame for my test case

obeleh:
Currently if I do this the `B` case will fail:

```python


class A:

    @typechecked
    def clone(self) -> 'A':
        return A()


val = A().clone()
assert isinstance(val, A)


def fn():

    class B:

        @typechecked
        def clone(self) -> 'B':
            return B()

    val = B().clone()
    assert isinstance(val, B)

fn()
```

The next commit fixes that but can you have a look?

obeleh:
Apparently I cannot use forward ref. Help!

Stewori:
Since you cannot know how deep the call is nested after the required type was declared, IMO you would have to combine various stack levels, e.g. like in [this util function](https://github.com/Stewori/pytypes/blob/master/pytypes/util.py#L694):

```
lcs = {}
for ln in stck[level:max_level]:
   lcs.update(ln[0].f_locals)
```

Going through the whole stack can be slow, so there should be some maximum number of layers that are searched, i.e. ``max_level``. Kind of best effort.
``level`` should probably be ``1`` here, it depends on whether typeguard calls this via another function call. If it's directly called from the decorator code, ``1`` should be the right choice.

I noticed that you check for ``str`` [here](https://github.com/agronholm/typeguard/pull/39/files#diff-de86ff4a007a4d2e83331804f54b4391R473). That might not be sufficient as forward refs may be nested, e.g. ``Tuple[int, str, 'B', A]``. Resolving this is not totally trivial, you might want to take a look at [pytypes.resolve_fw_decl](https://github.com/Stewori/pytypes/blob/master/pytypes/type_util.py#L909).
(Disclaimer: May be broken for Python 3.7 atm)

Looking at the code again, it appears to me that you attempt to add support for forward references and fix locals/globals regarding ``get_type_hints``. That's two separate things, none of them simple. At least I'd recommend them to be separate PRs if not separate issues at all.

typeguard.py
```diff
@@ -482,9 +499,13 @@ def typechecked(func: Callable = None, *, always: bool = False):
         warn('no type annotations present -- not typechecking {}'.format(function_name(func)))
         return func
 
+    frame = inspect.stack()[1][0]
+
+    dec_later = _inject_forward_refs(frame, func)
```
Stewori: This looks strange to me for two resons.

*  ``dec_later`` appears not to be used anywhere else
* ``_inject_forward_refs`` does not return anything

obeleh: sry copy pasta :(

agronholm:
If the `B` instance is passed to another function, how is typeguard supposed to find the stack frame where `B` was defined? It won't exist anymore!

Stewori:
Maybe @typechecked could resolve forward declarations at definition time?

agronholm:
> Maybe @typechecked could resolve forward declarations at definition time?

How? The class in @obeleh 's example doesn't yet exist at that point.

agronholm:
I've committed a fix which in part uses the same approach. I've credit you for your work even though I did not merge this PR.

PR #42, opened 2018-08-10
Fix for typing.* types not having a __name__

### LINKED ISSUE #41: typing.* types do not have __name__ and they error when generating messages.
```
        T = TypeVar('T', int, Collection)
        def foo(a: T, b: T):
            assert check_argument_types()

        with pytest.raises(TypeError):
            foo('aa', 'bb')
```

Shouldn't fail but does since `Collection.__name__` does not exist.

dalejung:
fixes #41


agronholm:
Thanks!

PR #44, opened 2018-09-11
Fix collections DeprecationWarning on python 3.7

tfiers:
The error in full (when typechecking for a Sequence):

```
typeguard.py:201: DeprecationWarning: Using or importing the ABCs from
'collections' instead of from 'collections.abc' is deprecated, and in
3.8 it will stop working

  if not isinstance(value, collections.Sequence):
```


agronholm:
Thanks!

PR #52, opened 2018-10-26
Special case Union[some_type, None], aka. Optional[some_type]

### LINKED ISSUE #51: Support common usage of Optional.
It seems, at least in my experience, that a common pattern for Optional arguments is to provide a default value of `None` and then, optionally, add the `Optional[]` qualifier. This results in the expected type of the argument being `Union[some_type, None]`. This can result in a weird `TypeError` though due to the way `Union` is handled for type checking. For example:

```python
@typechecked
def func(arg: Optional[List[str]] = None):
    pass
```
This function if used like this:
```python
func(['value', None])
```
results in this `TypeError`:
```python
TypeError: type of argument "arg" must be one of (List, NoneType); got list instead
```
This doesn't really indicate the actual reason this passed list doesn't match the type requirements of func.

Given this pattern of having the actual expected type, or None indicating it was not passed, seems common I propose special casing the handling of `Union[some_type, None]`. The actual value of the argument can be checked against `None` and if it isn't None then the proper type check against `some_type` is performed. Otherwise if it is `None`, then no further checking needs to happen.

tot0:
fixes #51

### DIFF
--- a/tests/test_typeguard.py
+++ b/tests/test_typeguard.py
@@ -4,7 +4,7 @@ from functools import wraps, partial
 from io import StringIO
 from typing import (
     Any, Callable, Dict, List, Set, Tuple, Union, TypeVar, Sequence, NamedTuple, Iterable,
-    Container, Generic)
+    Container, Generic, Optional)
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tot0:
@agronholm

tot0:
@agronholm Would really appreciate a review! Thanks!

agronholm:
Will take a look today.

agronholm:
Damn, forgot again. Will try again today or tomorrow.

agronholm:
I've taken a look at it now but I don't like it. Issue #51 stems from the fact that errors raised due to a mismatched `Union` can be hard to figure out and this necessitates a more robust error reporting mechanism that tells the user why each of the elements in the union failed to match the type in question. Special casing `Option` like this seems like a dirty hack to me.

tot0:
I agree that `Union` probably needs a wider fix that can maybe keep track of the various errors and present them in a more useful way. Though a `Union` of more than two types that aren't `None` makes this a very hard problem as far as giving a user useful information goes. Could easily result in wayyy too much info, most of which isn't useful.

I thought special casing for `Optional[]` use wasn't too hacky since `Optional[type]` is a very common usage of type hints and currently the error message in this case can be very cryptic/unhelpful. The fix is also scoped to only handle the case of `Union[type, None]`, this is a very specific (and common, imo) use case where the user only needs to know if what they passed didn't match `type`. The small code impact and brevity of the fix is what made me think it was worth doing by itself, especially given the use case it gives better support to. I do not have the time to come up with a more general fix (if there's even a good one for Union).

If you still don't wish to accept the fix that's unfortunate, though fair enough.

agronholm:
I'll need to ponder this some more. Generally though I would prefer to keep new features going to [pytypes](https://pypi.org/project/pytypes/).

agronholm:
Static type checkers don't special case `Optional`, so I've decided not to add that to Typeguard either.

PR #59, opened 2019-04-10
Make type hints cache thread-safe

kelseyfrancis:
When the same function is type-checked in multiple threads, a race
condition exists between the full initialization of that function's type
hints in one or more threads and iterating over those hints in one or
more other threads.

This can result in `RuntimeError: OrderedDict mutated during iteration`
in functions that iterate `_CallMemo.type_hints` items.

This fix is to only add the fully initialized type hints dict to the
global type hints cache rather than put an empty dict in the cache and
then subsequently populate that dict.

### DIFF
--- a/typeguard.py
+++ b/typeguard.py
@@ -45,7 +45,7 @@ class _CallMemo:
         self.type_hints = _type_hints_map.get(func)
         if self.type_hints is None:
             hints = get_type_hints(func)
-            self.type_hints = _type_hints_map[func] = OrderedDict()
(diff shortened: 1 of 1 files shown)


agronholm:
Seems reasonable. Thanks!

RichardJTorres:
@agronholm Wondering when you might be able to cut a release for this? We've hit this as well.

agronholm:
I've now pushed out the v2.3.1 release.

kelseyfrancis:
Thanks!

PR #63, opened 2019-07-12
Speed up getting caller's frame

pirate:
While not officially supported, `sys._getframe()` is waaay faster than `inspect.currentframe()` (on the order of 5x faster).

For benchmarks, see: https://gist.github.com/JettJones/c236494013f22723c1822126df944b12#gistcomment-2962311

This PR attempts to use the `sys._getframe()` if available, and falls back to the slower `inspect` method if not.
Because I don't know of any python implementation that doesn't support `sys._getframe()`, I don't actually know what `Exception` type would be thrown in case of it being unsupported, so for now I just used `Exception`.

(Note: `sys._getframe(1)` is the same as `sys._getframe(0).f_back`)

### DIFF
--- a/typeguard/__init__.py
+++ b/typeguard/__init__.py
@@ -472,7 +472,7 @@ def check_argument_types(memo: Optional[_CallMemo] = None) -> bool:
         try:
             # much faster than inspect, but not officially
             # supported in all python implementations
-            frame = sys._getframe(0).f_back
+            frame = sys._getframe(1)
         except Exception:
             # fall back to slower, but safer method
             frame = inspect.currentframe().f_back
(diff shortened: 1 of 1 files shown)


agronholm:
In what actual case might `sys._getframe(0)` raise an exception, and what exception would that be?

pirate:
https://bugs.python.org/issue6712

agronholm:
Ok, I understand now. Seems like IronPython is the only implementation without `sys._getframe()`. But even if you fall back to `inspect.currentframe()`, it will just return `None` in such a case, which will crash typeguard anyway. I think we'll cross that bridge when we come to it. In the meantime, you can drop the fallback.

pirate:
done @agronholm

agronholm:
Thanks!

PR #69, opened 2019-08-07
[TypeChecker] suppress false positive in the exception case

### LINKED ISSUE #68: `TypeChecker` reports return type mismatch when exception is raised
I currently use typeguard's `TypeChecker` to non-intrusively test one of my python projects and it works quite nicely in general. However, I noticed that exceptions seem to trip up the checking. I reduced the problem to the following example:

~~~python3
from typeguard import TypeChecker


def foo(arg: int) -> int:
    if arg > 10:
        raise Exception()
    return 42


checker = TypeChecker(__name__)
checker.start()

foo(5)
foo(12)
~~~
Executing this snippet gives me, besides the exception, the following return type mismatch warning: `TypeWarning: [MainThread] return from __main__.foo() at tgdemo.py:6: type of the return value must be int; got NoneType instead`

Is this a bug in the `TypeChecker`? The `typechecked` decorator at least does not complain about mismatching return values when an exception is raised.

niosHD:
Unfortunately, exceptions and regular returns of None seem to be
indistinguishable via the profiler interface. As the result,
this change also hides real warnings.

For now I only documented this limitation. However, better ideas to deal with this issue are more than welcome.

fixes #68

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,6 +31,7 @@ There are three principal ways to use type checking, each with its pros and cons
    * noninvasive (only records type violations; does not raise exceptions)
    * does not work reliably with dynamically defined type hints (e.g. in nested functions)
    * may cause problems with badly behaving debuggers or profilers
+   * can not distinguish returning None from raising an exception (i.e. detects slightly less)
 
 If a function is called with incompatible argument types or a ``@typechecked`` decorated function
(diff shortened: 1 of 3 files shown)


agronholm:
I made my own patch for this, but thanks anyway.

PR #76, opened 2019-08-28
Fixed yielded Iterable return types to not error

MSeal:
The latest 2.5.0 release caused

```
    @typechecked
    def yielding_func(self) -> Iterable[str]:
      yield 'foo'
```

to start failing with

```
typeguard/__init__.py:706: in wrapper
    return TypeCheckedGenerator(retval, memo)
typeguard/__init__.py:599: in __init__
    memo.type_hints['return'].__args__
E   ValueError: not enough values to unpack (expected 3, got 1)
```

Adding a catch for the return hints being a length 1 tuple for Iterator class yields (as opposed to 3 length Generator tuples) fixed the issue.

### DIFF
--- a/tests/test_typeguard.py
+++ b/tests/test_typeguard.py
@@ -903,6 +903,18 @@ class TestTypeChecked:
 
         assert exc.value.value == ['2', '3', '4']
 
+    def test_iterable_yield(self):
+        @typechecked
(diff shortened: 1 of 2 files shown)


agronholm:
This needs a broader fix. I have the changes underway.

jolaf:
#77 may be related.

PR #78, opened 2019-09-13
do not raise TypeError(TypeError(...)) exceptions

russok:


agronholm:
I feel dumb for not having spotted this before. Thanks!

PR #93, opened 2019-11-13
Fix crash when warning about annotated functions if function is object or partial function.

jtbeach:
- function_name() is called when we warn about unannotated functions in the @typechecked decorator.
- Inside the typechecked decorator, `func` can be any callable, not just functions
- `function_name` was relying on being called only with functions, but was actually therefore potentially being called with any unannotated callable
- Two instances of situations where `__qualname__` is not available for callables: partially-applied functions + objects with `__call__` methods.

### DIFF
--- a/tests/test_typeguard.py
+++ b/tests/test_typeguard.py
@@ -1022,6 +1022,24 @@ class TestTypeChecked:
         assert Child.foo('bar') == 'Child'
         pytest.raises(TypeError, Child.foo, 1)
 
+    def test_decorator_factory_no_annotations(self):
+        class CallableClass:
+            def __call__(self):
+                pass
+
+        def decorator_factory():
+            def decorator(f):
+                cmd = CallableClass()
+                return cmd
+
+            return decorator
+
+        with pytest.warns(UserWarning):
+            @typechecked
+            @decorator_factory()
+            def foo():
+                pass
+
 
 class TestTypeChecker:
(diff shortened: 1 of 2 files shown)


agronholm:
Please add a test that fails without your fix.

agronholm:
How are you doing type checking anyway against such things?

jtbeach:
> How are you doing type checking anyway against such things?

```python
import click

@typechecked
@click.command()
def cli():
    pass
```

Produces the following output:

```
  File "test_click.py", line 6, in <module>
    @click.command()
  File "/Users/jtbeach/.virtualenvs/blaze-api-py3.7/lib/python3.7/site-packages/typeguard/__init__.py", line 728, in typechecked
    warn('no type annotations present -- not typechecking {}'.format(function_name(func)))
  File "/Users/jtbeach/.virtualenvs/blaze-api-py3.7/lib/python3.7/site-packages/typeguard/__init__.py", line 201, in function_name
    qualname = func.__qualname__
AttributeError: 'Command' object has no attribute '__qualname__'
```

This is because the `click.command` decorator is returning a callable object.

agronholm:
Interestingly, if `@functools.wraps()` is used to decorate `decorator`, the test passes straight away. I'm guessing Click doesn't bother with that.

agronholm:
Thanks!

PR #105, opened 2020-01-03
Add support for Mock and MagicMock types.

prescod:
These are analogous to Any, except for values instead of arguments.

Any accepts any value. MagicMock can pretend to be any type.

### DIFF
--- a/tests/test_typeguard.py
+++ b/tests/test_typeguard.py
@@ -4,6 +4,7 @@ import warnings
 from concurrent.futures import ThreadPoolExecutor
 from functools import wraps, partial, lru_cache
 from io import StringIO, BytesIO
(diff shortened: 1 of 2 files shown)


prescod:
Hey @agronholm : are you able to take a look at this PR? It's fairly small and you agreed in principle to the approach back in December.

agronholm:
Thanks!

KantiCodes:
Hey,

Should that work?

```python
def test_something():
    import typeguard

    class B:
        pass
    
    @typeguard.typechecked
    class A:
        def __init__(self, b: B):
            self.b = b
    
    a = A(B())
    assert isinstance(a, A)


    a_mocked = A(MagicMock())
    ```

PR #109, opened 2020-02-02
Fixing PEP 561 support

xcthulhu:


xcthulhu:
Hmm... nevermind.  I thought this was an issue but it appears I just had an old copy of this module.

PR #110, opened 2020-02-02
Fixing PEP 561 support

xcthulhu:


xcthulhu:
This PR fixes PEP 561 support.  All I had to do was fix how `setup.cfg` was handling `package_data`.  Prior to this fix I was running into issues with using this package with `mypy`.  I've tested this locally and `mypy` should work now.

agronholm:
Uh, setuptools's own documentation says it's `options.package_data`: https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

agronholm:
I just tested and `pip install typeguard` does install `py.typed` in the right place.

xcthulhu:
You are right... it appears I was working with an older version!  Sorry for the confusion.

PR #114, opened 2020-03-02
Support backport Literal from typing_extensions

rfrowe:
Allows the type checker to support `Literal` types in python `3.5` through `3.7` if `typing_extensions` is installed.

For example: I'm writing a program targeted to python 3.8 but must maintain support for 3.7. The official [typing_extensions](https://github.com/python/typing/tree/master/typing_extensions) package backports `Literal`. However, despite this type now being available in my 3.7 runtime, typeguard's type checking fails to raise TypeError on invalid values.

Unfortunately, it's a little complicated. In `3.5 <= py < 3.7`, the `Literal` type does not have either `__origin__` nor `__args__`, so this is special cased in `check_type`

### DIFF
--- a/typeguard/__init__.py
+++ b/typeguard/__init__.py
@@ -29,7 +29,7 @@ except ImportError:
 try:
     from typing_extensions import Literal as BPLiteral
 except ImportError:
-    pass
+    BPLiteral = None
 
 try:
     from typing import AsyncGenerator
--- a/typeguard/__init__.py
+++ b/typeguard/__init__.py
@@ -27,6 +27,11 @@ except ImportError:
     Literal = TypedDict = None
 
 try:
+    from typing_extensions import Literal as BPLiteral
+except ModuleNotFoundError:
+    pass
+
+try:
     from typing import AsyncGenerator
 except ImportError:
     AsyncGenerator = None
@@ -465,9 +470,15 @@ def check_typevar(argname: str, value, typevar: TypeVar, memo: Optional[_CallMem
 
 
 def check_literal(argname: str, value, expected_type, memo: Optional[_CallMemo]):
-    if value not in expected_type.__args__:
+    try:
+        args = expected_type.__args__
+    except AttributeError:
+        # Instance of Literal from typing_extensions.
+        args = expected_type.__values__
+
+    if value not in args:
         raise TypeError('the value of {} must be one of {}; got {} instead'.
-                        format(argname, expected_type.__args__, value))
+                        format(argname, args, value))
 
 
 def check_number(argname: str, value, expected_type):
@@ -523,6 +534,8 @@ if Type is not None:
     origin_type_checkers[Type] = check_class
 if Literal is not None:
     origin_type_checkers[Literal] = check_literal
+if BPLiteral is not None:
+    origin_type_checkers[BPLiteral] = check_literal
 
 generator_origin_types = (Generator, collections.abc.Generator,
                           Iterator, collections.abc.Iterator,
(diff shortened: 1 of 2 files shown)


rfrowe:
@agronholm This (and all PRs) is blocked until #127 is merged, though it's ready for review again and passes locally.

I ran into this issue again during my work. `check_type('', n, Union[float, Literal['constant']])` should only succeed for `n: float = ...` or `n: Literal['constant'] = 'constant`, but in fact it succeeds for all `n: Any = ...`

typeguard/__init__.py
```diff
@@ -26,6 +26,11 @@
 except ImportError:
     Literal = TypedDict = None
 
+try:
+    from typing_extensions import Literal as BPLiteral
```
agronholm, line 30: What's the "BP" for?

typeguard/__init__.py
```diff
@@ -26,6 +26,11 @@
...
+except ModuleNotFoundError:
```
agronholm, line 31: How did you get the tests to pass locally on Python 3.5? ModuleNotFoundError was only introduced in 3.6.

typeguard/__init__.py
```diff
@@ -465,9 +470,15 @@ def check_typevar(argname: str, value, typevar: TypeVar, memo: Optional[_CallMem
 
 
 def check_literal(argname: str, value, expected_type, memo: Optional[_CallMemo]):
-    if value not in expected_type.__args__:
+    try:
+        args = expected_type.__args__
+    except AttributeError:
+        # Instance of Literal from typing_extensions.
```
agronholm, line 476: Small nit: drop the period at the end of the comment for consistency.

typeguard/__init__.py
```diff
@@ -595,6 +608,9 @@ def check_type(argname: str, value, expected_type, memo: Optional[_CallMemo] = N
     elif isinstance(expected_type, TypeVar):
         # Only happens on < 3.6
         check_typevar(argname, value, expected_type, memo)
+    elif BPLiteral is not None and isinstance(expected_type, BPLiteral.__class__):
+        # Only happens on < 3.7 when using Literal from typing_extensions
```
agronholm, line 612: Presumably happens on any Python version when using Literal from typing_extensions, no?

typeguard/__init__.py
agronholm: BackPort?

rfrowe: Yes. It can be made more explicit. I believe a different name is required later for conditional accessing the back port's literal values.

typeguard/__init__.py
rfrowe: Heh, tested with 3.8, 3.7, and 3.6. I didn't have 3.5 and the CI tests pass because `typing_extensions` is indeed installed. Would you recommend `except ImportError` then?

typeguard/__init__.py
rfrowe: `typing_extensions` is smart and will return the standard `Literal` implementation if it exists.

```python3
# Python 3.8.0 (default, Oct 28 2019, 16:14:01)
# [GCC 8.3.0] on linux
# Type "help", "copyright", "credits" or "license" for more information.
>>> from typing import Literal
>>> from typing_extensions import Literal as BPLiteral
>>> assert Literal is BPLiteral
>>>
```

In this case, the origin type checker dict lookup will yield `check_literal` and this code path will not be taken.

rfrowe: On Python 3.7, the implementation of `Literal` is different in `typing_extensions` than in Python 3.6. Namely, the generics typing API is used, so in this case the origin type checker dict lookup will again yield `check_literal`. This is why I've added `check_literal` to the dict under `BPLiteral` as well.

typeguard/__init__.py
agronholm: Right, of course. Yes, use `ImportError` here please.

typeguard/__init__.py
agronholm: This is okay, I was just curious.

typeguard/__init__.py
agronholm: Ok, got it.

agronholm:
Thanks!

PR #118, opened 2020-03-24
Don't try to wrap generators that are untyped.

prescod:
Generators that do not have return type signatures set would cause a crash within typeguard. This fixes that.

### DIFF
--- a/typeguard/__init__.py
+++ b/typeguard/__init__.py
@@ -814,7 +814,7 @@ def typechecked(func=None, *, always=False, _localns: Optional[Dict[str, Any]] =
         if inspect.isgenerator(retval) or isasyncgen(retval):
             return_type = memo.type_hints.get('return')
             if return_type:
-                origin = getattr(return_type, '__origin__')
+                origin = getattr(return_type, '__origin__', None)
                 if origin in generator_origin_types:
                     return TypeCheckedGenerator(retval, memo)
                 elif origin is not None and origin in asyncgen_origin_types:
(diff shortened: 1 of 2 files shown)


typeguard/__init__.py
```diff
@@ -813,11 +813,12 @@ def wrapper(*args, **kwargs):
         # If a generator is returned, wrap it if its yield/send/return types can be checked
         if inspect.isgenerator(retval) or isasyncgen(retval):
             return_type = memo.type_hints.get('return')
-            origin = getattr(return_type, '__origin__')
-            if origin in generator_origin_types:
-                return TypeCheckedGenerator(retval, memo)
-            elif origin is not None and origin in asyncgen_origin_types:
-                return TypeCheckedAsyncGenerator(retval, memo)
+            if return_type:
```
agronholm: What if `return_type` is `None`?

prescod: I think it would have to be "NoneType" rather than "None".

agronholm: In Python 3, `NoneType` is not importable. PEP 484 states that `None` is used in its place, so functions returning `None` will have `None` as the return type.

alexmojaki:
```python
import typing


def foo() -> None:
    pass


print(typing.get_type_hints(foo))  # prints {'return': <class 'NoneType'>}
```

So @prescod is right that the check should work correctly.

typeguard/__init__.py
```diff
@@ -813,11 +813,12 @@ def wrapper(*args, **kwargs):
...
+                origin = getattr(return_type, '__origin__')
```
alexmojaki, line 817:
```suggestion
                origin = getattr(return_type, '__origin__', None)
```

As an alternative to the change in this PR, I think this is the line that needs fixing. Currently it's exactly equivalent to `return_type.__origin__`, I think the author just mixed up `getattr` and `dict.get` as we all do sometimes.

prescod: @alexmojaki : If you want to replace the code in the PR or open another one, that's cool with me. Thanks.

agronholm: You are correct, the missing `None` here was a mistake.

agronholm:
Thanks! And sorry for taking so long with this.

PR #125, opened 2020-05-09
Compare types of module attributes with their annotations after module import

Tiendil:
Hi!

What are you thinking about checking types of module-level attributes after module imported?

It will help to find errors like:

```
x: int = 1
y: int = x / 2 # type error here 
```

And it matches the approach of run-time type checking.

### DIFF
--- a/tests/dummymodule_3_6.py
+++ b/tests/dummymodule_3_6.py
@@ -1,0 +1,13 @@
+
+# module for syntax that does not supported in Python 3.5
+
+import os
+
+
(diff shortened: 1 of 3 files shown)


agronholm:
I'm not leaning towards this because it might surprise people that it only works on the module level and not class/function level. Also, mypy usually does a pretty good job catching these so there isn't that much value in having typeguard check these.

Tiendil:
You have a point here. I will think about alternative variants. Sorry for the delayed response.

PR #132, opened 2020-06-03
Prevent all SpecialForms from using check_literal

### LINKED ISSUE #131: Functions with NoReturn return types fail to type check
## Reproduction
```python3
>>> from typing import NoReturn
>>> from typeguard import typechecked

>>> @typechecked
>>> def foo() -> NoReturn:
>>>    pass

>>> foo()
# AttributeError: '_SpecialForm' object has no attribute '__values__'
```

## Expected Behavior 

Passes type checking. (Although, maybe, it should fail because `NoReturn` shouldn't return but that's another conversation).

## Cause

#114

rfrowe:
Fixes #131 

@agronholm Sorry to introduce a bug like this. Wasn't caught by the existing test suites and I didn't catch it until testing my project with typeguard==2.8.0. I've added a regression test. Would appreciate a patch release :)

### DIFF
--- a/tests/test_typeguard_py38.py
+++ b/tests/test_typeguard_py38.py
@@ -1,4 +1,4 @@
-from typing import Literal, TypedDict, Union
+from typing import Literal, NoReturn, TypedDict, Union
 
 import pytest
 
@@ -22,6 +22,14 @@ def test_literal_union():
     foo(6)
     pytest.raises(TypeError, foo, 4).\
         match(r'must be one of \(str, typing.Literal\[1, 6, 8\]\); got int instead$')
+
+
+def test_regression_131():
+    @typechecked
+    def foo() -> NoReturn:
+        pass
+
+    foo()
 
 
 @pytest.mark.parametrize('value, total, error_re', [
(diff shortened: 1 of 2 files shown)


agronholm:
Closing this because supporting `NoReturn` fixed the referenced issue.

rfrowe:
I'm hesitant to agree. `_SpecialForm` is the base class of many builtin types. It is true that supporting `NoReturn` may fix this one repro case, but if any class derived from `_SpecialForm` is not supported by typeguard (say, a new builtin type is added in a new Python version), it will use `check_literal` without this PR

agronholm:
Yeah, but I don't see how this PR would fix that situation either.

rfrowe:
The condition `BPLiteral is not None and Literal is None` will evaluate to `False` in Python < 3.8 because `typing` has no `Literal`.

`BPLiteral.__class__ == typing._SpecialForm` is `True` only in Python 3.8.

Therefore, the last condition `isinstance(expected_type, BPLiteral.__class__)` will never be evaluated with `BPLiteral.__class__ == typing._SpecialForm`, so this fixes the situation I described.

PR #141, opened 2020-08-14
Update README.rst

wbolster:
counting is hard :wink:


agronholm:
Seems like I forgot to update that word after adding the fourth method. Thanks.

agronholm:
As they say, there are just three really hard problems in computing: cache invalidation and off-by-one errors.

PR #142, opened 2020-08-14
Use correct function name in documentation

wbolster:
It should be check_return_type(), but the example used check_return_value().


agronholm:
Oops. Thanks again!

PR #156, opened 2020-10-22
Fix #155

### LINKED ISSUE #155: typechecked decorator erases return type of decorated function
Test file:

```python

@typechecked
def foo(x: int) -> int:
    return bar(x)

@typechecked
def bar(x: int) -> int:
    return x
```

mypy 0.790 screams about returning `bar(x)` from foo:
```
[bash]$ mypy --strict --show-error-codes issue155.py
issue155.py:5: error: Returning Any from function declared to return "int"  [no-any-return]
Found 1 error in 1 file (checked 1 source file)
```

kstauffer:
Maybe I'm not understanding mypy or the typing module entirely, but it seems that TypeVar(name, bound=Callable[....]) is not treated the same as TypeVar(name, Callable[....], OtherType). But Type[T] is considered a subtype of Callable[..., T], at least in Python 3.8, and that subtype hierarchy is all typeguard needs.

### DIFF
--- a/typeguard/__init__.py
+++ b/typeguard/__init__.py
@@ -65,7 +65,7 @@ _type_hints_map = WeakKeyDictionary()  # type: Dict[FunctionType, Dict[str, Any]
 _functions_map = WeakValueDictionary()  # type: Dict[CodeType, FunctionType]
 _missing = object()
 
-T_CallableOrType = TypeVar('T_CallableOrType', Callable, Type[Any])
+T_CallableOrType = TypeVar('T_CallableOrType', bound=Callable[..., Any])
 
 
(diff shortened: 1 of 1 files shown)


kstauffer:
The section [Declaring decorators](https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators) in the mypy docs uses exactly this construct (`F = TypeVar("F", bound=Callable[..., Any])` and `def decorator(func: F) -> F`) as the canonical way to annotate the type of a decorator, so I think this is the recommended approach.

agronholm:
Thanks!

kstauffer:
Happy to help! Thanks for merging!

PR #158, opened 2020-10-23
Fix docs warning

EPronovost:
Fixes a doc warning that's causing the table to be missing from here https://typeguard.readthedocs.io/en/latest/userguide.html#supported-typing-types

Also adds a tox env to easily build docs.

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -210,7 +210,7 @@ Type            Notes
                 more information); On Python 3.9.0 or ``typing_extensions``
                 <= 3.7.4.3, false positives can happen when constructing
                 ``TypedDict`` classes using old-style syntax (see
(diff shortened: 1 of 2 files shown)


agronholm:
I don't think the docs env is necessary. I mean, `pip install -e .[doc] && python setup.py build_sphinx` should be easy enough, yes?

EPronovost:
Yeah that works! I wasn't sure how to build the docs.  What would you think of adding a `CONTRIBUTING.md` page to capture information like this?

agronholm:
Thanks. I don't see a contribution documentation as a priority right now though. Sphinx is by far the most popular tool for generating Python documentation, and I have not had anyone asking how to use it to generate documentation. Maybe some time in the future I'll add such instructions.

PR #166, opened 2020-12-22
fix: only check yielded and sent values to generators when annotated by user

muggenhor:
Without this we end up trying to enforce that every value yielded from
an iterator has the same type as those coming before it.

The added test case illustrates this.


agronholm:
The provided test seems to pass on current master, so doesn't look like this PR is necessary?

agronholm:
As per my comment from last year, I'm closing this. Let me know if you feel that was in error.

Issue #65, opened 2019-07-14
Make error messages more informative when a Union check fails


agronholm:
When an argument does not match a `Union`, the error message does not provide any hints as to why it specifically did not match each of the elements. This ought to be fixed to make troubleshooting easier.

agronholm:
I don't have the energy to cram this into the next release, as it requires significant effort.

jsh9:
I took a look at your source code (here: https://github.com/agronholm/typeguard/blob/1faec9d930c2b0f9bde841525703a35e8de834c1/typeguard/__init__.py#L408), and I think a quick fix may be to change Line 408 into:
```python
typelist = ', '.join(str(t) for t in union_params)
```

Then `typelist` looks like this:
```python
typing.Dict[int, float], typing.List[float, int]
```

(I admit that having `typing.` showing may not be the most elegant way. )

However, there may be another way: the `__args__` attribute of `t` is a tuple that looks like this: `(<class 'float'>, <class 'int'>)`. So maybe we can extract `"float"` and `"int"` and construct a nicer looking `typelist` string?

I can submit a pull request if you think my solutions are feasible.

agronholm:
I don't understand at all what you're getting at. Could you elaborate a bit? Also, this is a problem that can't really be solved properly with a change of a single line. If it was, I would've done it already.

jsh9:
Sorry I didn't provide enough context.

Below is the script that I was using:
```python
import typeguard
from typing import Union, Dict

b = {'a': 2, 'b': 2.3}

typeguard.check_type('b', b, Union[Dict[float, str], Dict[int, float]])
```

And the error message is:
`TypeError: type of b must be one of (Dict, Dict); got dict instead`

==============================
My solution can potentially solve the ambiguity in this part of the error message: "must be one of (Dict, Dict)", but probably not the second part.

But I noticed that the error message from this line https://github.com/agronholm/typeguard/blob/1faec9d930c2b0f9bde841525703a35e8de834c1/typeguard/__init__.py#L403 is quite informative. So maybe you can collect them in each iteration, and then show them at this line? https://github.com/agronholm/typeguard/blob/1faec9d930c2b0f9bde841525703a35e8de834c1/typeguard/__init__.py#L409

Then the error message would look like this:
```
type of keys of b must be either float or int; got str instead
type of keys of b must be int; got str instead
type of b must be one of (typing.Dict[int, float], typing.List[float, int]), ...
```

[...]

PR #173, opened 2021-02-06
Print complete `typing` type names

gkivanov:
This partially addresses the existing problem with type erasure while checking Union Types reported in https://github.com/agronholm/typeguard/issues/65

Currently, for this code example:
```
from typing import Optional, List

@typechecked
def myfun(arg: Optional[List[int]]) -> List[int]:
    return arg or []

myfun(["a"])
```

`typeguard` returns the following error, which is not very descriptive and can be quite ambiguous to users of the library being developed:
```
TypeError: type of argument "arg" must be one of (List, NoneType); got list instead
```

After the proposed change in this PR it returns the more descriptive error:
```
TypeError: type of argument "arg" must be one of (typing.List[int], NoneType); got list instead
```

Even though this does not completely resolve the Union issue in https://github.com/agronholm/typeguard/issues/65, it significantly reduces the ambiguity caused by it.


gkivanov:
@agronholm I will appreciate your comments

agronholm:
This is already fixed in master, so I'm closing the PR.

PR #180, opened 2021-03-16
Add decorator to disable run time type checking

### LINKED ISSUE #179: Decorator to exclude functions from runtime type checks
Would you accept a PR adding a decorator that excludes functions from runtime type checks with Typeguard, while still allowing for static type checks?

**Is your feature request related to a problem? Please describe.**

My primary use of Typeguard is as a mandatory check in CI using the pytest plugin. Typeguard has repeatedly identified bugs that were invisible to static type checkers. Unfortunately, this also means that I regularly run into problems due to discrepancies between static and runtime type-checking. Here are some examples from Typeguard's issue tracker:

- #178
- #138
- #115

We briefly talked about this before, in #138.

**Describe the solution you'd like**

Here's a POC:

```python
from typing import TypeVar
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    F = TypeVar("F")
    def typeguard_ignore(f: F) -> F:
        return f
else:
    from typing import no_type_check as typeguard_ignore

@typeguard_ignore
def f(arg: int) -> int:
    return arg

def test_f() -> None:
    f("teapot")
```

This code passes at runtime:

```sh
$ pytest --typeguard-packages=repro repro.py
...
1 passed in 0.01s
```

...but it's still caught by static type checkers:

```sh
$ mypy repro.py
repro.py:19: error: Argument 1 to "f" has incompatible type "str"; expected "int"
```

**Describe alternatives you've considered**

- Defining `typeguard_ignore` downstream (quite happy to do this if you prefer)
- Using `@no_type_check` (loses static type checks)
- Finding a work around for the special case (not always possible)

cjolowicz:
Closes #179

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -14,6 +14,8 @@ API reference
 
 .. autodecorator:: typechecked
 
+.. autodecorator:: typeguard_ignore
+
 .. autoclass:: TypeChecker
 
 .. autoexception:: TypeHintWarning
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -153,6 +153,17 @@ You can also customize the logic used to select which modules to instrument::
 
     install_import_hook('', cls=CustomFinder)
 
+To exclude specific functions or classes from run time type checking, use the ``@typeguard_ignore`` decorator::
+
+    from typeguard import typeguard_ignore
+
+    @typeguard_ignore
+    def f(x: int) -> int:
+        return str(x)
+
+Unlike :func:`~typing.no_type_check`, this decorator has no effect on static type checking.
+
+
 Using the pytest plugin
 -----------------------
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**UNRELEASED**
+
+- Added ``@typeguard_ignore`` decorator to exclude specific functions and classes from
+  runtime type checking (PR by Claudio Jolowicz)
+
 **2.11.1 (2021-02-16)**
 
 - Fixed compatibility with Python 3.10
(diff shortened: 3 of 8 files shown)


agronholm:
Looking good so far. I'll do a more thorough pass this weekend and merge if I see no issues.

agronholm (changes requested):
Ok, this is an elegant solution! If you could also add a changelog entry, I'll merge right away then.

agronholm (approved):

agronholm:
Thanks a lot! This was perfect work.

cjolowicz:
Thank you for merging this @agronholm :)

JeroenBos:
Thank you for implementing this decorator. I was keen to remove my workaround when updating `typeguard`, however, I realize that this decorator is not as versatile as I had hoped.

I was hoping that this decorator could be useful e.g. to prevent a `TypeError` from being raised when testing that certain errors are thrown, e.g.
```python
import unittest
class TestErrors(unittest.TestCase):
    def test_that_arg_of_wrong_type_raises(self):
        with typeguard_ignore(some_method):  # or ignore_typeguard_on from below
            with self.assertRaises(ValueError):
                some_method(some_wrong_value)

```

The idea being that normally typeguards are in effect, except under certain test conditions.

For full information, also to others that might have the same question, this is the workaround I had:

```python
from contextlib import contextmanager
from typing import Any
from unittest.mock import patch
from weakref import WeakKeyDictionary

@contextmanager
def ignore_typeguard_on(any: Any):
    """
    A helper to disable typeguard on the specified object,
    for when you explicitly want to pass in wrongly typed objects.
    """

    # disable a cache employed by typeguard:
    with patch("typeguard._type_hints_map", WeakKeyDictionary()):

        # set all annotations temporarily to typing.Any
        annotations = getattr(any, "__annotations__", None)
        if annotations is not None:
            with patch.dict(annotations, {key: Any for key in annotations.keys()}):
                yield
        else:
            yield
```

Issue #184, opened 2021-04-16
QUESTION: dealing with Numpy


luk-f-a:
**Is your feature request related to a problem? Please describe.**
I work extensively with numpy, and have annotated many functions as `def foo(x: int)`. However, I get many errors because typeguard considers `np.int32` or `np.int64` not to be `int`. 

```python
from typeguard import typechecked
import numpy as np

@typechecked
def foo(x: int):
    return x

arr = np.ones(1, dtype='i8')
foo(arr[0])
```

Is there an easy way to deal with this, besides creating a type alias for the union of int, int32, int64 and changing all the annotations?

Thanks!

agronholm:
Not at the moment. A hypothetical plugin (if typeguard had plugins) could address this.

luk-f-a:
thanks @agronholm , then I'll proceed with the Union.

PR #185, opened 2021-04-22
Add ability to override typeguard default checks

luk-f-a:
This PR is related to the question here: https://github.com/agronholm/typeguard/issues/184

After following the path of adding `Union[int, np.int64]` everywhere, which fixed the problem with typeguard, now I have 30 issues with mypy because `np.int64` behaves as an int (so it can be passed to `range` for example) but mypy will complain if a variable which is typed `Union[int, np.int64]` is passed to `range`.

This PR allows the user to provide a function that, for example, makes typeguard treat `np.int64` as `int`.

[x] added function register_override
[x] added tests
[x] extended docs

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -27,6 +27,7 @@ API reference
 
 .. autofunction:: register_override
 
+.. autofunction:: remove_override
 
 :mod:`typeguard.importhook`
 ---------------------------
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -25,6 +25,9 @@ API reference
 .. autoclass:: ForwardRefPolicy
     :members:
 
+.. autofunction:: register_override
+
+
 :mod:`typeguard.importhook`
 ---------------------------
 
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -230,3 +230,27 @@ Type            Notes
 
 .. _#101: https://github.com/agronholm/typeguard/issues/101
 .. _issue 42059: https://bugs.python.org/issue42059
+
+Overriding default checking
+---------------------------
+
+Occasionally, a project might need to override typeguard's default checking
+behaviour for a specific type.
+
+The `register_override` function provides the user the ability to submit a
+function that performs the type check for a given type.
+The override is type specific, subtypes of that type will not be affected.
+
+Overriding type checking can be necessary when working with libraries that
+rely on duck typing, without making that behaviour explicit in their
+type annotations. For example, Numpy has several integer types (e.g. `int64`)
+which are treated as equivalent to Python's built-in `int`. However, using
+them as equivalent will lead to typeguard raising an error about `int64` and
+`int` not being the same type.
+
+Example: overriding `int` checking to accept numpy integers.
+.. code-block:: python
+
+   def check_int(argname: str, value: Any, expected_type: Any, memo: Any):
+       if expected_type is int and not isinstance(value, (int, np.signedinteger)):
+           raise TypeError('type of {} must be either python int or numpy int: got {} instead'.format(argname, value.__class__))
--- a/src/typeguard/__init__.py
+++ b/src/typeguard/__init__.py
@@ -99,7 +99,8 @@ def register_override(expected_type: Type,
 
        def check_int(argname: str, value: Any, expected_type: Any, memo: Any):
            if expected_type is int and not isinstance(value, (int, np.signedinteger)):
-               raise TypeError('type of {} must be either python int or numpy int: got {} instead'.format(argname, value.__class__))
+               raise TypeError('type of {} must be either python or numpy int: got {} instead'
+               .format(argname, value.__class__))
 
     """
     _overrides[expected_type] = check_function
--- a/src/typeguard/__init__.py
+++ b/src/typeguard/__init__.py
@@ -75,7 +75,7 @@ else:
 _type_hints_map = WeakKeyDictionary()  # type: Dict[FunctionType, Dict[str, Any]]
 _functions_map = WeakValueDictionary()  # type: Dict[CodeType, FunctionType]
 _missing = object()
-_overrides = {} # type: Dict[Type, Callable[[str, Any, Any, Any], None]]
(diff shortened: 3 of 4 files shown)


luk-f-a:
@agronholm the PR is ready for your review. I am open to discussing the concept, mechanism and UX of the override idea. This PR is a complete implementation of the idea but also starting point for discussion if you don't like something in particular.
The tests on commit 694b3db did not pass, but I cannot reproduce the failure locally. I created a fresh local environment, checked that python and all main packages are exactly the same version as in CI console, but still cannot reproduce the failure. I don't understand how my change could have triggered that failure.
The motivating example for this PR is the following:
```python
import numpy as np
from typing import Union, List

arr = np.ones(1, dtype=np.int64)*5

@typechecked
def foo(arg: int) -> List[int]:
    return list(range(arg))

print(type(arr[0])) # <class 'numpy.int64'>
foo(arr[0])
```
The code above works well, because np.int64 is compatible with python `int`. However, typeguard would complain since it's not a subtype of `int`.
The natural annotation improvement is to use a `Union` but then this happens:

```python
import numpy as np
from typing import Union, List

arr = np.ones(1, dtype=np.int64)*5

@typechecked
def foo(arg: Union[int, np.signedinteger]) -> List[int]:
    return list(range(arg))

print(foo(arr[0]))
```
In this case, mypy complains with `Argument 1 to "range" has incompatible type "Union[int, signedinteger[Any]]";
expected "int"`

It would be possible to fill the code with `int(arg)` but in a large codebase where many such cases, it makes the code less readable.
I think the override mechanism is good way to solve this, being flexible while not affecting users that don't want to use it.

luk-f-a:
update: I have managed to reproduce the error locally. I hope to have a fix soon.

luk-f-a:
update: all tests are passing. PR is ready.

agronholm:
My plan was to have a broader configuration system which could be applied in a granular fashion. I'll keep this PR open until I figure that out. It will be part of the eventual 3.0 release then.

luk-f-a:
I can't say I wasn't hoping this would be merged, but I understand that if you're planning something broader, then you'd rather wait for that. I hope this can serve as a checkpoint for a use case that would ideally be covered in that configuration system.

agronholm:
Sure. If you (or anybody else) are willing to take up maintainership of this project, I'd be willing to give it up because I don't have much time to work on this anymore.

luk-f-a:
for personal reasons I can't take on a project unfortunately.

luk-f-a:
@agronholm , sorry to bother you but I wonder what my options are. You prefer not to merge this PR, but it's also unlikely that 3.0 will happen in the short term.

I need this PR but I cannot pip install from my github fork. I could vendor it (not ideal) or I could completely fork the project under a new name and publish to PyPI (which would be cleaner). Would it bother you if I published my fork under a different name? You'll still get full attribution of course.

agronholm:
I won't have time for typeguard 3.0 for a while, so let's explore your options.
You said you can't install your github fork. Would you like to elaborate on that? What is preventing you from doing that?
If there is a legitimate reason, I won't (or can't) stop you from forking the project. It is the least preferred option though because that increases the clutter on PyPI.

luk-f-a:
yeah, I agree duplicating on PyPI is not great. I can't install from github fork due to the corporate firewall. This needs to be installed by different people and also CI agents. Each of them under different firewall rules. Conda install and pip install are fine, but not github.
Is it hard to build a wheel and keep a local copy from which to pip install? I thought about just copying the whole code in a sub-folder of the application, but I don't know what'll happen with dependencies and the pytest plugin.

agronholm:
`pip wheel` should give you the dependencies to install from locally.

agronholm:
There is now ongoing work to support hooks for specifying external type checkers in the 3.0 branch.

agronholm:
The 3.0 branch contains a way to register plugins for this purpose. Let me know if it doesn't suit your needs.

luk-f-a:
@agronholm thanks for pinging me. I had a look and I have two questions on the 3.0 feature:
- the example in the docs refers to a custom class. What about replacing the type checking for an existing type, even a built-in like `int`? Would this feature allow that?
- to register the plugin it says that `After modifying your project configuration, you may have to reinstall it in order for the entry point to become discoverable.`. What does it mean having to "reinstall it"? I use typeguard for the development of my own package, do I need to install my own package to get this plugin to work? Isn't there a more "stateless" way to register the plugin (like in here https://github.com/luk-f-a/typeguard/blob/6dc70b3fb5b6489bc21728fdc0cdde18b92fdcfa/src/typeguard/__init__.py#L83-L84)?

agronholm:
> the example in the docs refers to a custom class. What about replacing the type checking for an existing type, even a built-in like int? Would this feature allow that?

Yes. Custom checker lookup functions would be called first, before falling back to the built-in ones.

> to register the plugin it says that After modifying your project configuration, you may have to reinstall it in order for the entry point to become discoverable.. What does it mean having to "reinstall it"? I use typeguard for the development of my own package, do I need to install my own package to get this plugin to work? Isn't there a more "stateless" way to register the plugin (like in here https://github.com/luk-f-a/typeguard/blob/6dc70b3fb5b6489bc21728fdc0cdde18b92fdcfa/src/typeguard/__init__.py#L83-L84)?

Entry points are the only workable way to advertise the existence of such plugins. Otherwise, how is typeguard going to be able to load all the installed plugins if it doesn't know about them?

luk-f-a:
> Otherwise, how is typeguard going to be able to load all the installed plugins if it doesn't know about them?

what about providing a function, like `register_override` in the link, that performs manual registration? automatic discovery is great but sometimes is necessary to register things manually.

agronholm:
A developer can already do `from typeguard import config; config.checker_lookup_functions.insert(your_lookup_function)`.

luk-f-a:
thanks for the explanation. may I suggest adding that to the documentation? It's useful, and it would avoid having to dig into the code to figure it out.

agronholm:
The whole `TypeCheckConfiguration` class is still waiting to be documented. This attribute will be documented along with the rest.

luk-f-a:
thanks for adding this to 3.0

PR #190, opened 2021-05-26
Fix `AttributeError` when `__code__` is missing from the checked callable

epenet:
Fixes: #191

When using https://github.com/pallets/click, the `@click.group()` function decorator return a class instance (`click.core.<Group xyz>`) which does not have associated code. This causes Typeguard to fail with `AttributeError: 'Group' object has no attribute '__code__'`.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,10 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**UNRELEASED**
+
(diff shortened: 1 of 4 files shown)


agronholm:
The test suite fails on Python 3.5 because of the variable annotation. It is on my to-do list to drop 3.5 support in this project but until then, move the test to `test_typeguard_py36.py`.

epenet:
@agronholm I have moved the tests

agronholm (approved):
Looks good!

agronholm:
Thanks!

epenet:
Thank you for merging this @agronholm :)

PR #202, opened 2021-09-15
fix(init) don't warn on unannotated properties

### LINKED ISSUE #194: Warnings with no declarations
**Describe the bug**

One of the reasons we prefer TypeGuard to mypy is because it only checks arguments that we actually specify the types for.

Recently, typeguard seems to warn (not error) on properties (but not functions) that do not have type declarations.

**To Reproduce**

Given this file, I get warnings for "baz" (which I do not like) but not for "bar" (which is great).

```
import typeguard

@typeguard.typechecked
class MyClass:
    def foo(self, x: int):
        pass

    def bar(self):
        pass

    @property
    def baz(self):
        pass

MyClass().foo(5)
```

Warning output:

```
/Users/pprescod/.venvs/CCI-37/lib/python3.9/site-packages/typeguard/__init__.py:1125: UserWarning: no type annotations present -- not typechecking __main__.MyClass.baz /Users/pprescod/code/CumulusCI/test.py:12
```

**Expected behavior**

This file has no runtime type errors and should generate no warnings.

**Additional context**

Changing the test to this fixes the problem:

```
    if getattr(func, "__annotations__", None) is None:
```

I'm not sure if this is the correct fix because I'm not sure what these warnings were supposed
to warn against.

https://github.com/agronholm/typeguard/blob/89c1478bd33bcf9a7cccc2c962ebeaa034e51908/src/typeguard/__init__.py#L909

@epenet , what do you think? It looks like you added this test.

AckslD:
closes #194

### DIFF
--- a/src/typeguard/__init__.py
+++ b/src/typeguard/__init__.py
@@ -906,7 +906,7 @@ def typechecked(func=None, *, always=False, _localns: Optional[Dict[str, Any]] =
 
         return func
 
-    if not getattr(func, '__annotations__', None):
+    if getattr(func, "__annotations__", None) is None:
         warn('no type annotations present -- not typechecking {}'.format(function_name(func)))
         return func
 


agronholm:
Help me understand this fix. The only thing it would change is the behavior when `__annotations__` is falsy but not `None`.

agronholm:
Also, a fix would generally require an added test that would fail without the fix.

AckslD:
@agronholm I just took the fix from @prescod in #194 since it also worked for me and put it in a PR

agronholm:
This fix is wrong. If the problem is decorating a class that has unannotated methods then the correct fix is to just skip those methods. It should still warn if the user tries to explicitly annotate functions that have no annotations.

AckslD:
@agronholm I'm afraid I won't have time at the moment to dig into the code but in that case I'll close this now.

Sorry for a not so worked-out-PR.

Btw, what do you mean with warning a user that tries to annotate a function with no annotations?

agronholm:
> Btw, what do you mean with warning a user that tries to annotate a function with no annotations?

Sorry, I meant decorating a function with no annotations.

AckslD:
Why would decorating an annotated function give a warning? This is what this fix was supposed to fix since if I understand the point of this package correctly then it should just ignore parts of the code which is not annotated so that one can gradually move to typing.

agronholm:
> Why would decorating an annotated function give a warning?

It wouldn't. It would give a warning for un-annotated functions, to tell the user that they're not supposed to do that.
This is only a problem when decorating classes instead of functions directly. In this case it should just skip over the methods that have no type annotations.

AckslD:
What is it that the user is not supposed to do exactly? Sorry for dragging this along, I'm not trying to be annoying, just to understand :)

For example, wouldn't the following be fine and shouldn't give any warnings?
```python
class SomeClass:
    @property
    def some_property_without_annotations():
        pass
```

agronholm:
What users are not supposed to do is:
```python3
@typechecked
def foo(a, b):
    pass
```

AckslD:
That makes sense 👍

PR #210, opened 2021-11-04
Fix __module__ default when run in an exec

andyljones:
Running typeguard 2.13 in an exec with an empty environment errors. This PR fixes that by mapping a `f.__module__` of `None` to the empty string.

```
exec("""

@typechecked
def f():
   pass

""", {})
```

(The root issue is that this exec-in-empty-namespace thing is what doctests like to do)


agronholm:
Thanks!

Issue #227, opened 2022-01-03
check_type issue with types.UnionType (PEP 604)


vbrinnel:
Thanks a lot for `typeguard`!
It seems `check_type` has issues with `types.UnionType` annotations:

```
from typeguard import check_type
from typing import Union
check_type("k", 1, Union[str, list]) # OK: "TypeError: type of k must be one of (str, list); got int instead"
check_type("k", 1, str | list) # Not OK: returns None
```

jeeftor:
But the latest version is only 2.3.x right?

agronholm:
> But the latest version is only 2.3.x right?

Yes, what's your point? 3.0 is not yet out due to outstanding issues and lack of time to devote to the project (as I have multiple projects requiring my attention).

PR #228, opened 2022-01-03
Fix https://github.com/agronholm/typeguard/issues/227

vbrinnel:
Adds `UnionType` support for `check_type`
```
check_type("k", 1, str|list) # OK -> TypeError: type of k must be one of (str, list); got int instead
```


agronholm:
Typeguard already supports UnionType in the 3.0 branch.

PR #243, opened 2022-03-02
Import is_typeddict from typing_extensions if possible

### LINKED ISSUE #242: typing_extensions.TypedDict not detected on Python 3.10
**Describe the bug**
typeguard doesn't work with `typing_extensions.TypedDict` on Python 3.10 with typing_extensions >= 4.1.

**To Reproduce**
Run tests:

    tox -e py310

which fails due to

    TypeError: TypedDict does not support instance and class checks

**Expected behavior**
typeguard recognizes `typing_extensions.TypedDict`s as TypedDicts and all tests succeed.

**Additional context**
Since typing_extensions 4.1 implements its own implementation of `TypedDict` to support PEP 655, so `typing.is_typeddict()` doesn't recognize `typing_extensions.TypedDict`.

d-k-bo:
Fixes #242.


agronholm:
All development happens in the 3.0 branch. Is this fix still relevant there?

agronholm:
Also, if this fixes a problem, you need to add a test that would fail without the fix.

d-k-bo:
> All development happens in the 3.0 branch. Is this fix still relevant there?

I wasn't aware of the 3.0 branch. The issue still persists, but needs to fixed in `_utils.py`.

> Also, if this fixes a problem, you need to add a test that would fail without the fix.

[This test](https://github.com/agronholm/typeguard/blob/de6ab051309ba74a0a27840f8172697c8778ae4f/tests/test_typeguard.py#L1333) on the master branch already fails, since it imports `TypedDict` from `typing_extensions`. I added a small that test to the new branch.

d-k-bo:
Sorry for all this git mess, I guess it would be better if a opened a new PR.

tests/test_checkers.py
```diff
@@ -391,6 +391,18 @@ class DummyDict(TypedDict, total=total):
         else:
             check_type(value, DummyDict)
 
+    @pytest.mark.skipif(
+        sys.version_info < (3, 10), reason="Only required on Python 3.10+"
+    )
```
agronholm, line 396: Does it actually hurt if this test is run on older Pythons?

agronholm:
Thanks!

tests/test_checkers.py
d-k-bo: No, but it would require `typing_extensions` as an additional test dependency on Python 3.9. I think it isn't required, because the problematic `is_typeddict` only exists since Python 3.10.

agronholm: Yup, I came to the same conclusion and decided to drop the issue.

d-k-bo:
Thanks for merging!

PR #247, opened 2022-03-31
Avoid no-any-return in `typeguard._exeptions.TypeCheckError.__str__`

eggplants:
`mypy --strict` says:

```
src/typeguard/_exceptions.py:33: error: Returning Any from function declared to
return "str"  [no-any-return]
                return " of ".join(self._path) + " " + self.args[0]
                ^
src/typeguard/_exceptions.py:35: error: Returning Any from function declared to
return "str"  [no-any-return]
                return self.args[0]
                ^
```

If `self.args` is an empty tuple, it raises `IndexError`. So I add:

```python
if len(self.args) == 0:
    raise ValueError("No args are given.")
```

### DIFF
--- a/src/typeguard/_exceptions.py
+++ b/src/typeguard/_exceptions.py
@@ -29,7 +29,9 @@ class TypeCheckError(Exception):
         self._path.append(element)
 
     def __str__(self) -> str:
-        if self._path:
-            return " of ".join(self._path) + " " + self.args[0]
+        if len(self.args) == 0:
+            raise ValueError("No args are given.")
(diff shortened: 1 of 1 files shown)


agronholm:
`self.args` can never be empty because of this: `super().__init__(message)`. There's just no way to tell mypy that, so ignoring that error with a type comment would be a better solution.

agronholm:
I've committed said fix to the 3.0 branch.

This fixes just 2 of 39 total mypy errors anyway. Do you intend to make this project mypy clean altogether?

PR #251, opened 2022-04-21
Handles type checks for Type[Protocol]

williamlw999-fb:
I noticed that typeguard wasn't able to correctly type check passed in classes when the type hint was a Type[Protocol]. You would receive an error like: `TypeError: Protocols with non-method members don't support issubclass()`. Using isinstance() in these situations would be able to correctly type check the passed in class. I also added a warning when a Protocol wasn't annotated with runtime_checkable to make it more explicit when typeguard was unable to perform type checking.

Tests: 
Added a data protocol to improve coverage
Added a test for checking Type[Protocol]

### DIFF
--- a/src/typeguard/_checkers.py
+++ b/src/typeguard/_checkers.py
@@ -376,7 +376,9 @@ def check_class(
         return
 
     expected_class = args[0]
-    if isinstance(expected_class, TypeVar):
+    if getattr(expected_class, "_is_protocol", False):
+        check_protocol(value, expected_class, (), memo)
+    elif isinstance(expected_class, TypeVar):
         check_typevar(value, expected_class, (), memo, subclass_check=True)
     elif get_origin(expected_class) is Union:
         errors: Dict[str, TypeCheckError] = {}
@@ -501,6 +503,12 @@ def check_protocol(
             raise TypeCheckError(
                 f"is not compatible with the {origin_type.__qualname__} protocol"
             )
+    else:
+        warnings.warn(
+            "Typeguard cannot check the {} protocol because it is a non-runtime protocol. "
+            "If you would like to type check this protocol, "
+            "please use @typing.runtime_checkable".format(origin_type.__qualname__)
+        )
 
 
 def check_byteslike(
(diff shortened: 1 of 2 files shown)


deathowl:
Hey @agronholm , could you take a look at this PR? As this would be needed in testslide.

agronholm:
The 2.x series (current master) has been frozen for new features; you should rebase this against the 3.0 branch.

agronholm:
I've merged the 3.0 branch to master now to prevent any further confusion.

williamlw999-fb:
@agronholm Thanks for helping out with the 3.0 rebase! I've updated my PR and made sure all signals are green. Let me know if there's anything else you'd like me to address for this PR

agronholm:
This PR confuses me. You seem to be dropping type checking support for static protocols. Why?

williamlw999-fb:
Static protocols were never supported by Typeguard. See line 501 of _checkers.py: `if getattr(origin_type, "_is_runtime_protocol", False):`. 

StaticProtocol is defined without `@runtime_checkable`, so it does not pass that if statement in `check_protocol()`, so typeguard doesn't type check it and always returns None. (See python docs on the runtime_checkable decorator https://docs.python.org/3/library/typing.html#typing.runtime_checkable). I've added a warning to make sure that users are aware when their protocol is not being type checked, and I updated the test to check for that warning. 

The main fix in the PR is the addition of check_protocol when handling check_class for cases where you have a type hint like `Type[Protocol]`.

williamlw999-fb:
To be more specific, I've made sure that the the type checker can detect not only that the instance`Foo()` is a `RuntimeProtocol`, *but also* that the class `Foo` is a `Type[RuntimeProtocol]`. 

There were no changes made to the how Typeguard handles `StaticProtocol` aside from warning the user that Python (and subsequently Typeguard) can't check protocols not decorated with `@runtime_checkable` at runtime.

williamlw999-fb:
Does this help clarify the PR?

agronholm (changes requested):
Looks good, just one tiny nit.

agronholm (approved):

agronholm:
Thanks!

PR #252, opened 2022-04-25
add option to importhook all packages for pytest-plugin

PatrikHlobil:
Hi @agronholm,

I stumbled upon your module last week and find it amazing. In this MR I would like to add a pytest-option to autoimport all modules inside a test.  Would you be interested in this feature? 

I also thought about additional options, e.g. to be able to only specify the excluded modules that should not be autoimported.

Best Patrik


agronholm:
Hi, it's been a while. I'm actively working on this project again for a while, and I came across your PR. Can you explain me the motivation behind it?

PatrikHlobil:
Hi @agronholm,

I just want the option that in the pytest run, all imported modules will be checked with typeguard. At least for my team, this would be the desired case, because I do not only want to check the correct types on my own code, but also on the third-party code that I am using.

Does that make sense for you?

agronholm:
What do you run your tests with? Pytest at least would be potentially affected by this, as typeguard could interfere with pytest's import hook which rewrites asserts in tests.

As for the special value, I'm thinking of `:all:`, used by pip for a similar purpose. The dot might imply that only the packages in the current directory would get instrumented.

PatrikHlobil:
I think `:all:` would indeed be a better option. yes, we are using pytest for our tests.

agronholm:
I implemented this a bit differently.

PR #254, opened 2022-04-30
Make @typechecked decorator works when @static/classmethod decorators are used any order

jacobpbrugh:
Prior to this PR, the @typechecked decorator only works when added after 
@staticmethod or @classmethod decorators due to a bug caused by the
obscure property that classmethod and staticmethod objects do not inherit
from type. 

Previously adding @typechecked before either of these decorators would
result in a warning and no typechecking. After this PR @typechecked will
work with classmethod and staticmethod functions regardless of 
decorator order.


agronholm:
Please add the appropriate test(s) to ensure that this will work correctly in the future.

agronholm:
Thanks for this! I added the tests myself because I needed this in the upcoming release.

agronholm:
Oops, accidentally pressed "revert".

PR #258, opened 2022-06-03
Add Required/NotRequired typing

fcarpentey75012:
Hello Team, 

I add Required/NotRequired from typing_extensions, doesn't change the comportement of typechecked when you use total=False, but when you use total=True, the code check if we use NotRequired and doesn't flag the keys like an required_keys. I hope this fork can be usefull for another user.

Example : 
```
from dataclasses import dataclass, field
from typing import *
from typing_extensions import NotRequired, Required 

@dataclass
class Fields():
    """Metadata for the request
    """
    asset: Required[List[str]] = field(default_factory=['X478'])
    type_asset: NotRequired[List[str]] = field(default_factory="")

@typechecked
def test_toto(params: TypedDict("params",Fields.__annotations__, total = True)):
    pass

test_toto(params = {"asset" : ["10"]})
test_toto(params = {"asset" : ["10"], "type_asset" : [10]})

```

### DIFF
--- a/src/typeguard/_checkers.py
+++ b/src/typeguard/_checkers.py
@@ -2,7 +2,6 @@ import collections.abc
 import inspect
 import sys
 import types
-import typing_extensions
 import warnings
 from enum import Enum
 from inspect import Parameter, isclass, isfunction
@@ -29,6 +28,8 @@ from typing import (
     TypeVar,
     Union,
 )
+
+import typing_extensions
 
 from ._exceptions import TypeCheckError, TypeHintWarning
 from ._memo import TypeCheckMemo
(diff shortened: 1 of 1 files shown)


agronholm:
New tests are required to match the new functionality.

src/typeguard/_checkers.py
```diff
@@ -29,6 +29,8 @@
     Union,
 )
 
+import typing_extensions
```
agronholm, line 32: You need to follow existing conventions. Add this conditional import to section after unconditional imports:
```suggestion
if sys.version_info >= (3, 11):
    from typing import NotRequired, Required
else:
    from typing import NotRequired, Required
```

agronholm: You marked this as unresolved even though you didn't make any changes, why?

agronholm:
Do you intend to finish this PR?

agronholm:
I checked and it looks like to me that `Required` and `NotRequired` work fine with Typeguard, just the way they were documented in the `typing` module.

PR #264, opened 2022-07-09
fixed some checks for ANY

fefe982:
Typeguard currently fails on checks of : 

* returning Any
* Mapping with key or value annotated as Any
* Type[Any]

maybe more.

For the previous two, it uses `type_check_internal`, which does not check `Any`. The program tries to check for `Any` before calling `type_check_intenal`, but fails to do so in all places. I think it may be better to check for `Any` inside `type_check_internal`, because `Any` should be checked on every call site of `type_check_internal`.

The last is handled by `check_class`. `check_class` handled `Type[Union[Any]]`, but not `Type[Any]`.

### DIFF
--- a/src/typeguard/_checkers.py
+++ b/src/typeguard/_checkers.py
@@ -376,7 +376,9 @@ def check_class(
         return
 
     expected_class = args[0]
-    if getattr(expected_class, "_is_protocol", False):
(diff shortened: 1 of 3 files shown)


agronholm:
LGTM. Thanks a lot!

PR #278, opened 2022-12-30
Fixed an exception when a dict has mixed bytes and str keys

biolds:
The following code demonstrates the problem that is fixed by this PR:

```python
from typing import TypedDict

class TestDict(TypedDict):
    a: str

@typechecked
def test() -> TestDict:
    return {
        'str': 'test',
        b'bytes': 'test'
    }

test()
```

```
$ ./test.py 
  File "/home/ldef/typeguard/src/./test.py", line 16, in <module>
    test()
  File "/home/ldef/typeguard/src/typeguard/__init__.py", line 285, in wrapper
    check_return_type(retval, memo)
  File "/home/ldef/typeguard/src/typeguard/__init__.py", line 172, in check_return_type
    check_type_internal(retval, memo.type_hints["return"], memo)
  File "/home/ldef/typeguard/src/typeguard/_checkers.py", line 568, in check_type_internal
  File "/home/ldef/typeguard/src/typeguard/_checkers.py", line 220, in check_typed_dict
    raise TypeCheckError(f"has unexpected extra key(s): {keys_formatted}")
typeguard._exceptions.TypeCheckError: the return value has unexpected extra key(s): "b'bytes'", "str"
```

### DIFF
--- a/src/typeguard/_checkers.py
+++ b/src/typeguard/_checkers.py
@@ -200,6 +200,12 @@ def check_mapping(
 def check_typed_dict(
     value: Any, origin_type: Any, args: Tuple[Any, ...], memo: TypeCheckMemo
 ) -> None:
+    def sorted_dict(value):
+        try:
+            return sorted(value)
+        except TypeError:
+            return value
+
     declared_keys = frozenset(origin_type.__annotations__)
     if hasattr(origin_type, "__required_keys__"):
         required_keys = origin_type.__required_keys__
@@ -209,12 +215,12 @@ def check_typed_dict(
     existing_keys = frozenset(value)
     extra_keys = existing_keys - declared_keys
     if extra_keys:
-        keys_formatted = ", ".join(f'"{key}"' for key in sorted(extra_keys))
+        keys_formatted = ", ".join(f'"{key}"' for key in sorted_dict(extra_keys))
         raise TypeCheckError(f"has unexpected extra key(s): {keys_formatted}")
 
     missing_keys = required_keys - existing_keys
     if missing_keys:
-        keys_formatted = ", ".join(f'"{key}"' for key in sorted(missing_keys))
+        keys_formatted = ", ".join(f'"{key}"' for key in sorted_dict(missing_keys))
         raise TypeCheckError(f"is missing required key(s): {keys_formatted}")
 
     for key, argtype in get_type_hints(origin_type).items():
--- a/tests/test_inline.py
+++ b/tests/test_inline.py
@@ -11,6 +11,11 @@ if sys.version_info >= (3, 9):
 else:
     from typing_extensions import Annotated
 
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
 
 class TestCheckArgumentTypes:
     def test_valid(self):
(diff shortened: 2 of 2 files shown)


agronholm:
Please add a regression test for this.

biolds:
Done

agronholm:
The test suite fails for Python 3.7. You need to import `TypedDict` from `typing_extensions` on that version.

biolds:
I've fixed it

src/typeguard/_checkers.py
```diff
@@ -209,12 +216,12 @@ def check_typed_dict(
     existing_keys = frozenset(value)
     extra_keys = existing_keys - declared_keys
     if extra_keys:
-        keys_formatted = ", ".join(f'"{key}"' for key in sorted(extra_keys))
+        keys_formatted = ", ".join(f'"{key}"' for key in _sorted_dict(extra_keys))
```
agronholm, line 219: I'm thinking that we don't need actually need to sort the keys. The order should be stable since Python 3.6. What if you dicth `_sorted_dict()` entirely?

biolds: The order is actually lost just before, because frozensets are used to find the `extra_keys`.

agronholm: How is the order lost with frozensets?

agronholm: Never mind, yeah, they're not like dicts which keep key order.

agronholm: It does seem that frozensets already sort the keys, so I'm not sure why I was sorting the keys in the first place. Thoughts?

biolds: So, this is confusing, but I think I have found the reason. :)
set / frozenset use the `hash` built-in function and sorts hashed values, but the `hash` function has a seed that changes when the interpreter is started. So when testing in an interactive shell, the order may be ordered alphabetically (that could occur randomly depending on the seed value), and this order is maintained when dealing with the same frozenset. But then, exiting the interactive shell and starting may yield a different ordering for the same values inside the frozenset.
This explains why in the unittest I had to put a regexp that matches both orders when the keys are not sorted.

biolds: And dict keys maintain insertion order since Python 3.7 (https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects, see the note of the `iter` method)

biolds: The hashing is described there: https://docs.python.org/3/reference/datamodel.html#object.__hash__

agronholm: Yeah, I commented earlier about dicts keeping insertion order (from Python 3.6 actually, but 3.7 made it official). But it seems that the sort order does not change over restarts of the interpreter, so the only purpose of the sorting seems to be to prettify the key list.

This is just for the looks, but would you please move the `_sorted_dict` function inside `check_typed_dict()`, as it's not used elsewhere? You can remove the leading underscore from the name too.

biolds: Done.

agronholm:
Ok, this is perfect. Just fix that one pre-commit failure (one line is too long) and I'll merge. This just missed the 3.0.0b2 release, but I'll include this in the next one.

biolds:
Should be good now

agronholm:
Thanks!

biolds:
Thanks to you for this project! :)

agronholm:
I simplified the code and test a bit, and made the sort stable by keying it on the `repr()` of each item.

PR #284, opened 2023-01-24
Fix union transformer to handle `Callable[..., Any]`

supersergiy:


agronholm:
I think this needs a test in `test_importhook` (and a corresponding annotated function in `dummymodule.py`.

supersergiy:
@agronholm How's this?

agronholm:
Looks good, thanks!

PR #289, opened 2023-01-28
Converted @typechecked to recompile the target function with type checks

agronholm:
We no longer wrap the callable, thus eliminating the extra stack frame.


src/typeguard/_decorators.py
```diff
@@ -31,6 +27,60 @@ def typeguard_ignore(f: _F) -> _F:
 T_CallableOrType = TypeVar("T_CallableOrType", bound=Callable[..., Any])
 
 
+def make_cell():
+    value = None
+    return (lambda: value).__closure__[0]
+
+
+def instrument(f: T_CallableOrType) -> Callable | str:
+    if not getattr(f, "__annotations__", None):
```
marcelroed, line 36: Why not use `hasattr`?

src/typeguard/_decorators.py
```diff
@@ -31,6 +27,60 @@ def typeguard_ignore(f: _F) -> _F:
...
+    module_code = compile(module_ast, module.__file__, "exec", dont_inherit=True)
+    new_code = module_code
+    for level, name in enumerate(target_path):
+        for const in new_code.co_consts:
+            if isinstance(const, CodeType):
+                if const.co_name == name:
+                    new_code = const
+                    break
+        else:
+            return "cannot find the target function in the AST"
```
marcelroed, line 58: Could create an error type and raise it here

src/typeguard/_decorators.py
marcelroed: I suppose this also catches the case where `__annotations__ is None`?

agronholm: > I suppose this also catches the case where `__annotations__ is None`?

Indeed.

agronholm:
Quick reminder: [the other PR](https://github.com/agronholm/typeguard/pull/294) has all this plus the checking of annotated variable assignments.

agronholm:
Found a bug: in the case of `@overload`, it instruments one of the overloads and not the implementation.

agronholm:
Fixed that in the other PR.

agronholm:
@marcelroed I'm going to have to move on with these changes, so I can't wait much longer for your feedback. Let me know if you need more time.

marcelroed:
Go for it! I looked through it and it looks great!

agronholm:
Great, thanks!

agronholm:
> Go for it! I looked through it and it looks great!

Does this mean you also looked at variable assignment checking?

marcelroed:
Yep, I added some comments now, but I couldn't find any issues without going very deep. To be honest, any `ast` code is a bit opaque without stepping through with a debugger.

PR #291, opened 2023-02-02
Replaced the PEP 604 union transformer with a simpler one

agronholm:
This also makes it possible to use the built-in collection types as generic.


supersergiy:
What a beautiful idea!

agronholm:
Good enough for merging, then? :smiley:

supersergiy:
Yeah, looks good. Congrats on figuring this out!

agronholm:
Thanks for the review!

PR #294, opened 2023-02-07
Check variable assignments

agronholm:
This type checks assignments to annotated local variables and arguments in functions.


agronholm:
This PR is now a mish-mash of the feature in question, plus all the changes I've made since then. Sorry about that.

agronholm:
Found a bug: PEP 604 unions in variable annotations are not transformed into `typing.Union` on Python < 3.10. Fixing.

src/typeguard/_functions.py
```diff
@@ -260,6 +260,25 @@ def check_yield_type(yieldval: T, memo: CallMemo) -> T:
     return yieldval
 
 
+def check_variable_assignment(
+    value: object, expected_annotations: dict[str, Any], memo: CallMemo
```
marcelroed, line 264: Why are you using the object type instead of Any?

src/typeguard/_transformer.py
```diff
@@ -429,3 +478,105 @@ def visit_Yield(self, node: Yield):
...
+            )
+            expected_types = Dict(keys=[Constant(node.target.id)], values=[annotation])
+            check_call = Call(
+                self._get_typeguard_import("check_variable_assignment"),
+                [operator_call, expected_types, self._memo.get_call_memo_name()],
+                [],
+            )
+            node = Assign(targets=[node.target], value=check_call)
+
+        return node
```
marcelroed, line 582: Looks good, but I would like docstrings to explain the changes on visit for readability, perhaps showing what transformation is made with syntax

src/typeguard/_functions.py
agronholm: To treat `value` as opaque. Making it an `object` makes access to any of its (non-dunder) attribute invalid. I probably should've done this in many other places too.

agronholm:
> Found a bug: PEP 604 unions in variable annotations are not transformed into `typing.Union` on Python < 3.10. Fixing.

Fixed.

src/typeguard/_transformer.py
agronholm: I've added some docstrings to methods that make changes to the nodes passing through them. Is this satisfactory or would you say it needs more?

marcelroed: Looks good! I'm certainly satisfied.

agronholm:
Ok, one step closer now :smiley:
I'm going to make another change that avoids checking names imported within an `if TYPE_CHECKING:` branch.

PR #314, opened 2023-03-15
Add sub heading

utix:
This allow to have link to specific change log


agronholm:
I had headings in my changelog in the past, but I got annoyed by them generating too many TOC entries in the documentation, so I got rid of them. Therefore I will politely reject this.

Issue #316, opened 2023-03-16
Global config introduced in 3.0.0 isn't threadsafe


antonagestam:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Typeguard version

3.0.1

### Python version

3.11.0

### What happened?

Hi there 👋

I'm the maintainer of [phantom-types](https://github.com/antonagestam/phantom-types), a project that relies heavily on typeguard. It's very nice to see all the new activity going on here! 👍

While looking into the changes for 3.0.0, I noticed the change in default strategy for items of collections. phantom-types relies on the original behavior, and would have to configure `ALL_ITEMS` whenever making calls to `check_type`. However, I noticed that the configuration is really just a globally shared state, with no facilities for thread safety. As phantom-types is a library that is used in many different contexts by its users, some of them are bound to be using threads.

Hence, the current implementation in 3.0.0 wouldn't really be safe to start using, as it'd be only a matter of time before users stumble upon this race condition when using threads (or asyncio).

Would you be open to somehow addressing this? I think ideally from my perspective, there would be a way to pass in config directly to the `check_type` call, instead of relying on the integrity of globally shared state. What do you think?

Again, typeguard really enables phantom-types, and I'm plenty grateful for this library!

### How can we reproduce the bug?

In this example, we expect `target` to never be called, because locally in `thread_a` we're setting to check `ALL_ITEMS` and then calling with an invalid list. Yet, because another thread is concurrently altering the global config, only the first item of the list is checked, and `target` is called.

```python
from concurrent.futures import ThreadPoolExecutor
import time

import pytest
import typeguard


@typeguard.typechecked
def target(arg: list[int]) -> int:
    # This should never be reached, because we're calling with invalid arguments.
    assert all(isinstance(item, int) for item in arg), "not all items are int"
    return 1


def test_typeguard_threadsafe():
    typeguard.config.collection_check_strategy = (
        typeguard.CollectionCheckStrategy.FIRST_ITEM
    )

    def thread_a():
        typeguard.config.collection_check_strategy = (
            typeguard.CollectionCheckStrategy.ALL_ITEMS
        )
        # Uncommenting this sleep makes the test pass, proving there's a race condition.
        time.sleep(.2)
        with pytest.raises(typeguard.TypeCheckError):
            target([1, "2"])

    def thread_b():
        time.sleep(.1)
        typeguard.config.collection_check_strategy = (
[...]

PR #317, opened 2023-03-16
Make check_type independent of global config

antonagestam:
Addresses #316.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,10 +5,14 @@ This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-v
 
 **UNRELEASED**
 
-- **BACKWARD INCOMPATIBLE** Made ``check_type()`` independent of ``typeguard.config``, it
-  now has the same default configuration values as ``TypeCheckConfiguration`` and is
-  given the ability to accept configuration options as keyword arguments.
+- **BACKWARD INCOMPATIBLE** ``check_type()`` no longer uses the global configuration.
+  It now uses the default configuration values, unless overridden with an explicit
+  ``config`` argument.
 - **BACKWARD INCOMPATIBLE** Removed ``CallMemo`` from the API
+- **BACKWARD INCOMPATIBLE** Required checkers to use the configuration from
+  ``memo.config``, rather than the global configuration
+- Added keyword arguments to ``@typechecked``, allowing users to override settings on a
+  per-function basis
 - Changed instrumentation to statically copy the function annotations to avoid having to
   look up the function object at run time
 - Added support for type checking against nonlocal classes defined within the same
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,9 @@ This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-v
 
 **UNRELEASED**
 
+- **BACKWARD INCOMPATIBLE** Made ``check_type()`` independent of ``typeguard.config``, it
+  now has the same default configuration values as ``TypeCheckConfiguration`` and is
+  given the ability to accept configuration options as keyword arguments.
 - **BACKWARD INCOMPATIBLE** Removed ``CallMemo`` from the API
 - Changed instrumentation to statically copy the function annotations to avoid having to
   look up the function object at run time
--- a/src/typeguard/_functions.py
+++ b/src/typeguard/_functions.py
@@ -8,7 +8,12 @@ from threading import Lock
 from typing import Any, Callable, Iterable, NoReturn, TypeVar, cast, overload
 
 from ._checkers import BINARY_MAGIC_METHODS, check_type_internal
-from ._config import global_config
+from ._config import (
+    CollectionCheckStrategy,
+    ForwardRefPolicy,
+    TypeCheckConfiguration,
+    global_config,
+)
 from ._exceptions import TypeCheckError, TypeCheckWarning
 from ._memo import TypeCheckMemo
 from ._utils import qualified_name
@@ -26,16 +31,41 @@ type_checks_suppress_lock = Lock()
 
 
 @overload
-def check_type(value: object, expected_type: type[T]) -> T:
+def check_type(
+    value: object,
+    expected_type: type[T],
+    *,
+    forward_ref_policy: ForwardRefPolicy = ...,
+    typecheck_fail_callback: TypeCheckFailCallback | None = ...,
+    collection_check_strategy: CollectionCheckStrategy = ...,
+) -> T:
     ...
 
 
 @overload
-def check_type(value: object, expected_type: Any) -> Any:
+def check_type(
+    value: object,
+    expected_type: Any,
+    *,
+    forward_ref_policy: ForwardRefPolicy = ...,
+    typecheck_fail_callback: TypeCheckFailCallback | None = ...,
+    collection_check_strategy: CollectionCheckStrategy = ...,
+) -> Any:
     ...
 
 
-def check_type(value: object, expected_type: Any) -> Any:
+def check_type(
+    value: object,
+    expected_type: Any,
+    *,
+    forward_ref_policy: ForwardRefPolicy = TypeCheckConfiguration().forward_ref_policy,
+    typecheck_fail_callback: (TypeCheckFailCallback | None) = (
+        TypeCheckConfiguration().typecheck_fail_callback
+    ),
+    collection_check_strategy: CollectionCheckStrategy = (
+        TypeCheckConfiguration().collection_check_strategy
+    ),
+) -> Any:
     """
     Ensure that ``value`` matches ``expected_type``.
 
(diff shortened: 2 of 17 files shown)


src/typeguard/_functions.py
```diff
@@ -71,18 +83,21 @@ def check_type(value: object, expected_type: Any) -> Any:
...
         else:
             raise
 
     return value
 
 
-def check_argument_types(memo: CallMemo) -> Literal[True]:
+def check_argument_types(
+    memo: CallMemo,
+    config: TypeCheckConfiguration = global_config,
```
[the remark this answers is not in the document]
antonagestam: > My own changes, which I scrapped earlier, contained the configuration in the memo.

Let me know which way you want to go, I have no strong opinion either way.

> What's the point of adding this parameter to these functions, when it's never being passed any explicit value?

I thought it made sense for these to be in-line with the signature of `check_type`. The version history says these functions are removed, should they be?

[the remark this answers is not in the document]
agronholm: > I thought it made sense for these to be in-line with the signature of check_type. The version history says these functions are removed, should they be?

They're only used internally now, by both `@typechecked` and the import hook.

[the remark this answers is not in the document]
antonagestam: I've restored these changes, addressing `@typechecked` and the import hook to not use global state could be done separately from this PR.

antonagestam:
@agronholm I addressed your comment about unnecessarily changing the internal functions, I agree that makes sense. I do think it will make sense to address the decorator and import hook in a future PR though, my opinion is that relying on a globally shared and mutable configuration is a recipe for unpleasant surprises, but again, that's opinion.

I made another change here in that I unpacked the configuration parameters into explicit arguments for `check_type`, instead of taking a single configuration object. This is also opinionated, my reasoning is that it's a bit simpler API, less work and import on the call sites of `check_type`. Let me know if you're not on board with this, and I'll happily roll back to taking a `TypeCheckConfiguration` there instead.

A third change, which makes this PR a breaking change, is that the new arguments of `check_type` now default to the _default values of `TypeCheckConfiguration`_, and __not to the values in `global_config`__. Again, this is my opinion about globally shared state and surprises, I just think it makes sense for the defaults of this function to be static.

Docs and tests are updated, and this PR is ready for a re-review.

src/typeguard/_functions.py
```diff
@@ -61,21 +91,32 @@ def check_type(value: object, expected_type: Any) -> Any:
...
+        see :attr:`TypeCheckConfiguration.collection_check_strategy`
     :return: ``value``, unmodified
     :raises TypeCheckError: if there is a type mismatch
 
     """
+    config = TypeCheckConfiguration(
+        forward_ref_policy=forward_ref_policy,
+        typecheck_fail_callback=typecheck_fail_callback,
+        collection_check_strategy=collection_check_strategy,
+    )
```
agronholm, line 107: This overrides all of the configuration, even if the user just wants to override a single option.

antonagestam: Not really, the defaults gathered from the defaults from of `TypeCheckConfiguration`: https://github.com/agronholm/typeguard/blob/2f4c104770adcc3b0d53bbc52a7f05bf787808b2/src/typeguard/_functions.py#L71

But the semantics are different from `@typechecked` and import hook, in that this is completely detached from the global config. My reasoning is that `check_type` is likely to be used differently from those, like in my case with phantom-types, and in this context it doesn't make sense to have shared mutable state. Does that make sense? :)

src/typeguard/_functions.py
```diff
@@ -36,16 +41,41 @@
...
+) -> Any:
     ...
 
 
-def check_type(value: object, expected_type: Any) -> Any:
+def check_type(
+    value: object,
+    expected_type: Any,
+    *,
+    forward_ref_policy: ForwardRefPolicy = TypeCheckConfiguration.forward_ref_policy,
```
antonagestam, line 71: Should probably create a default instance of `TypeCheckConfiguration` instead of referencing class attributes, because this would most likely break if a definition is using `dataclasses.field()`.

src/typeguard/_functions.py
agronholm: > Not really, the defaults gathered from the defaults from of `TypeCheckConfiguration`:

The issue is that it disregards the current global config entirely, unlike how it works in `master`. This is one breaking change I'm not sure I want to merge.

antonagestam: That's correct, as I mentioned in the comment above, I want to challenge the choice to make this function's behavior depend on shared mutable state. I understand why it makes sense for `@typechecked` and the import hook to have a global config object, but I think it's unwise for this function to depend on it.

Consider if these field are made to default to the global config, in order to fulfill the case where you don't want to be affected by the globally shared state, then you would always have to pass all configuration. What's worse, if typeguard adds more configuration options to this function in the future, it would be very hard for a user to detect that these have been added and that they should update all their calls to include a static default value.

I.e., my opinion is that `check_type` is sufficiently different from the instrumenting APIs to be treated differently.

agronholm: Alright, you're starting to convince me. This change would have to be reflected in the documentation too then. And `@typechecked` needs to have the overrides added to it too.

src/typeguard/_functions.py
agronholm: Maybe `forward_ref_policy: ForwardRefPolicy = TypeCheckConfiguration().forward_ref_policy` ?

docs/versionhistory.rst
```diff
@@ -8,6 +8,9 @@ This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-v
 - Fixed ``warn_on_error()`` not showing where the type violation actually occurred
 - Fixed local assignment to ``*args`` or ``**kwargs`` being type checked incorrectly
 - Fixed ``TypeError`` on ``check_type(..., None)``
+- **BACKWARD INCOMPATBILE** Made ``check_type`` independent of ``typeguard.config``, it
+  now has the same default configuration values as ``TypeCheckConfiguration`` and is
+  given the ability to accept configuration options as keyword arguments.
```
agronholm, line 13: Backward incompatible changes should be displayed first.
```suggestion
- **BACKWARD INCOMPATIBLE** Made ``check_type()`` independent of ``typeguard.config``, it
  now has the same default configuration values as ``TypeCheckConfiguration`` and is
  given the ability to accept configuration options as keyword arguments.
- Fixed ``warn_on_error()`` not showing where the type violation actually occurred
- Fixed local assignment to ``*args`` or ``**kwargs`` being type checked incorrectly
- Fixed ``TypeError`` on ``check_type(..., None)``
```

src/typeguard/_functions.py
antonagestam: Linters complained about instantiations in defaults, so I did almost as suggested but broke out to a module-local variable: `_default_config`.

docs/versionhistory.rst
antonagestam: Fixed.

docs/versionhistory.rst
```diff
@@ -5,12 +5,12 @@ This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-v
 
 **UNRELEASED**
 
-- Fixed ``warn_on_error()`` not showing where the type violation actually occurred
-- Fixed local assignment to ``*args`` or ``**kwargs`` being type checked incorrectly
-- Fixed ``TypeError`` on ``check_type(..., None)``
 - **BACKWARD INCOMPATBILE** Made ``check_type`` independent of ``typeguard.config``, it
```
agronholm, line 8: Fixed a typo and added () to indicate function.
```suggestion
- **BACKWARD INCOMPATIBLE** Made ``check_type()`` independent of ``typeguard.config``, it
```

src/typeguard/_functions.py
agronholm: What was the complaint?

docs/versionhistory.rst
antonagestam: This exceeds line-length, I'll push a re-formatting :)

antonagestam: Oh, noticed rest of file doesn't adhere to line-length limit, so I'll leave it :)

agronholm:
It looks like you need to do a merge from `master` before this PR becomes mergeable.

src/typeguard/_functions.py
antonagestam: For this diff:

```diff
diff --git a/src/typeguard/_functions.py b/src/typeguard/_functions.py
index db51c1e..66cc92e 100644
--- a/src/typeguard/_functions.py
+++ b/src/typeguard/_functions.py
@@ -73,7 +73,7 @@ def check_type(
     *,
     forward_ref_policy: ForwardRefPolicy = _default_config.forward_ref_policy,
     typecheck_fail_callback: (TypeCheckFailCallback | None) = (
-        _default_config.typecheck_fail_callback
+        TypeCheckConfiguration().typecheck_fail_callback
     ),
     collection_check_strategy: CollectionCheckStrategy = (
         _default_config.collection_check_strategy
```

I get this:

```sh
$ pre-commit run ruff --all-files
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1

src/typeguard/_functions.py:76:9: B008 Do not perform function call `TypeCheckConfiguration` in argument defaults
Found 1 error.
```

agronholm: That seems to come from a flake8-bugbear rule. It's safe to ignore. You can change add `"B008"` to `tool.ruff.ignore` in `pyproject.toml` where it already has `"PGH001"`.

antonagestam:
@agronholm Is Github being weird somehow? I see this:

Skärmavbild 2023-03-19 kl  15 11 52

And git reports being up-to-date with master (I synced yesterday, merge commit: [2f4c104](https://github.com/agronholm/typeguard/pull/317/commits/2f4c104770adcc3b0d53bbc52a7f05bf787808b2)):

```sh
$ git fetch -p upstream && git merge upstream/master 
Already up to date.
```

Do you want me to squash, or do you use Github's "Squash and merge" feature?

agronholm:
This is what I see:
image

> Do you want me to squash, or do you use Github's "Squash and merge" feature?

I always squash when merging PRs.

agronholm:
I now have a new branch for the upcoming v4.0: `4.0-dev`. This PR should target that branch.

agronholm:
There's a new checker that needs updating.

agronholm:
The failures on py3.7 came from my side. I'll fix them.

antonagestam:
I've squashed and rebased on the 4.0 branch, fixed conflicts, and reversed to prefer multiple instantiations of `TypeCheckConfiguration` over a shared instance for defaults, ignoring B008.

agronholm:
Sounds great! I'll fix the 3.7 failures either today or tomorrow. Meanwhile, you can add the new parameters to `typechecked()`. The AST transformer code (which also needs to be updated to support this) might look a bit scary, but I can handle that.

agronholm:
Do you want me to take over from here?

antonagestam:
@agronholm Please go ahead ☺️

agronholm:
I'm not yet done with this PR, but this latest changeset should show what I intend to do.

agronholm:
Thanks for this. I hope I can release v4.0 soon now.

antonagestam:
@agronholm Thanks for bringing this across the finish line 🙏

PR #321, opened 2023-03-23
Update README.rst import hook example to be compatible with 3.0.0

Danie-1:
Due to the backward incompatible change in 3.0.0, the current example for installing the import hook no longer works.


agronholm:
Damn, missed that one. Thanks!

PR #328, opened 2023-04-01
Call pytest indirectly

mtelka:
For better interoperability with `tox-current-env`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,7 @@ minversion = 4.0
 
 [testenv]
 extras = test
-commands = pytest {posargs}
+commands = python -m pytest {posargs}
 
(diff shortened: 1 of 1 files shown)


agronholm:
Could you explain this?

mtelka:
I'm packaging [typeguard for OpenIndiana](https://github.com/OpenIndiana/oi-userland/commits/oi/hipster/components/python/typeguard) and for testing we use the `tox-current-env` plugin.

Quote from https://pypi.org/project/tox-current-env/:

> tests should invoke `python -m pytest` rather than assuming the `pytest` command is present and uses the correct version of Python

agronholm:
Ok, not a problem.

Issue #352, opened 2023-05-05
`ValueError: Empty body on If` when defining class under conditional


rhaps0dy:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Typeguard version

3.0.2

### Python version

3.10.10

### What happened?

The following code causes a crash on import/run. 

```python
from typeguard import typechecked

if True:
    class A:
        ...

@typechecked
def f(a: A):
    ...
```

Prints:

```
Traceback (most recent call last):
  File ".../repro.py", line 8, in <module>
    def f(a: A):
  File ".../python3.10/site-packages/typeguard/_decorators.py", line 175, in typechecked
    retval = instrument(target)
  File ".../python3.10/site-packages/typeguard/_decorators.py", line 75, in instrument
    module_code = compile(module_ast, f.__code__.co_filename, "exec", dont_inherit=True)
ValueError: empty body on If
```

<details><summary>And indeed the AST contains an empty If statement. Click to expand.

```python
print(ast.dump(module_ast, indent=4))
```
</summary>

```python
Module(
    body=[
        ImportFrom(
            module='typeguard',
            names=[
                alias(name='typechecked')],
            level=0),
        If(
            test=Constant(value=True),
            body=[],
            orelse=[]),
        FunctionDef(
            name='f',
            args=arguments(
                posonlyargs=[],
                args=[
                    arg(
                        arg='a',
                        annotation=Name(id='A', ctx=Load()))],
                kwonlyargs=[],
                kw_defaults=[],
                defaults=[]),
            body=[
                ImportFrom(
                    module='typeguard',
                    names=[
                        alias(name='CallMemo')],
                    level=0),
                ImportFrom(
                    module='typeguard._functions',
                    names=[
                        alias(name='check_argument_types')],
                    level=0),
                Assign(
                    targets=[
                        Name(id='call_memo', ctx=Store())],
                    value=Call(
                        func=Name(id='CallMemo', ctx=Load()),
                        args=[
                            Name(id='f', ctx=Load()),
                            Call(
                                func=Name(id='locals', ctx=Load()),
                                args=[],
                                keywords=[])],
                        keywords=[])),
                Expr(
                    value=Call(
                        func=Name(id='check_argument_types', ctx=Load()),
                        args=[
                            Name(id='call_memo', ctx=Load())],
                        keywords=[])),
                Expr(
                    value=Constant(value=Ellipsis))],
[...]

PR #386, opened 2023-08-26
Add `pass` to removed blocks of code

vthemelis:
Solves #352 more generally.

We saw this in exception handlers but I think that it can manifest itself on any construct that has an expected child block.

We could potentially even totally remove the statements in the test as they are no-op.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .idea
 .tox
 .coverage
+**/.coverage.*
 .cache
 .eggs/
 *.egg-info/
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 .pydevproject
 .idea
 .tox
-.coverage
 .coverage*
 .cache
 .eggs/
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 .idea
 .tox
 .coverage
-**/.coverage.*
+.coverage*
 .cache
 .eggs/
 *.egg-info/
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,10 +7,14 @@ This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-v
 
 - Dropped Python 3.7 support
 
+- Fixed ``@typechecked`` optimization causing compilation of instrumented code to fail
+  when any block was left empty by the AST transformer (eg `if` or `try` / `except` blocks)
+  (`#352 <https://github.com/agronholm/typeguard/issues/352>`_)
+
 **4.1.2** (2023-08-18)
 
 - Fixed ``Any`` being removed from a subscript that still contains other elements
-  (`#373 <https://github.com/agronholm/typeguard/issues/373>`_
+  (`#373 <https://github.com/agronholm/typeguard/issues/373>`_)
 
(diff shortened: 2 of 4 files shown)


agronholm:
You need to fix the linting error and add a note to the changelog. What's with the gitignore change?

vthemelis:
> You need to fix the linting error

Fixed

>  and add a note to the changelog.

Added

> What's with the gitignore change?

I was getting a lot of `.coverage.<blah> ` files when running `tox` locally. I can remove the change if you prefer but those files are not covered by any of the existing rules and I don't think that they should ever be committed to source control.

.gitignore
```diff
@@ -3,6 +3,7 @@
 .idea
 .tox
 .coverage
+**/.coverage.*
```
agronholm, line 6:
```suggestion
.coverage*
```

docs/versionhistory.rst
```diff
@@ -7,10 +7,14 @@ This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-v
 
 - Dropped Python 3.7 support
 
+- Fixed ``@typechecked`` optimization causing compilation of instrumented code to fail
+  when any block was left empty by the AST transformer (eg `if` or `try` / `except` blocks)
+  (`#352 <https://github.com/agronholm/typeguard/issues/352>`_)
```
agronholm, line 12:
```suggestion
- Fixed ``@typechecked`` optimization causing compilation of instrumented code to fail
  when any block was left empty by the AST transformer (eg `if` or `try` / `except` blocks)
  (`#352 <https://github.com/agronholm/typeguard/issues/352>`_)
```

src/typeguard/_transformer.py
```diff
@@ -65,6 +66,8 @@
 from dataclasses import dataclass, field
 from typing import Any, ClassVar, cast, overload
 
+from typing_extensions import override
```
agronholm, line 69: Can't have an unconditional import of `typing_extensions` when it's a conditional dependency in `pyproject.toml`. Please remove the `@override` decorators. They're not essential to this change. I'll add them later if I deem them useful.

agronholm: Ignore; I removed it myself.

src/typeguard/_transformer.py
```diff
@@ -498,6 +499,21 @@ def __init__(
...
+        non_empty_list_fields = []
+        for field_name, val in iter_fields(node):
+            if isinstance(val, list) and len(val) > 0:
+                non_empty_list_fields.append(field_name)
+
+        node = super().generic_visit(node)
+
+        # Add `pass` to list fields that were optimised away
+        for field_name in non_empty_list_fields:
+            if not hasattr(node, field_name) or not getattr(node, field_name):
```
agronholm, line 512:
```suggestion
            if not getattr(node, field_name, None):
```

agronholm:
The test you added seems to already pass in `master`.

agronholm:
Thanks!

vthemelis:
@agronholm, I opened https://github.com/python/cpython/issues/108521 where they pointed out that maybe we should only be filling in `body` attributes of `AST` nodes. Would you like me to raise another PR to fix this?

agronholm:
Sure.

Issue #385, opened 2023-08-25
"SyntaxError: from __future__ imports must occur at the beginning of the file" after comments


jolaf:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Typeguard version

4.1.2

### Python version

3.11.2

### What happened?

```
Traceback (most recent call last):
  File "C:\Temp\test.py", line 4, in <module>
    import_module('A')
  File "C:\Python311\Lib\importlib\__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1206, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1149, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "C:\Python311\Lib\site-packages\typeguard\_importhook.py", line 98, in exec_module
    super().exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 936, in exec_module
  File "<frozen importlib._bootstrap_external>", line 1074, in get_code
  File "C:\Python311\Lib\site-packages\typeguard\_importhook.py", line 87, in source_to_code
    return _call_with_frames_removed(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\typeguard\_importhook.py", line 47, in _call_with_frames_removed
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "C:\Temp\A.py", line 4
    from __future__ import annotations
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: from __future__ imports must occur at the beginning of the file
```

### How can we reproduce the bug?

`test.py`:
```python3
from importlib import import_module
from typeguard import install_import_hook
with install_import_hook(('A',)):
    import_module('A')
```

`A.py`:
```python3
#
# Comment
#
from __future__ import annotations

def f() -> int:
    return 5

print("OK")
```
Run:

```shell
$ python3 A.py
OK

$ python3 test.py
<crash>
```

jolaf:
Thanks!!

PR #387, opened 2023-08-26
Fix imports in a module containing __future__

vthemelis:
Fixes: https://github.com/agronholm/typeguard/issues/385

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,12 +18,13 @@ repos:
     hooks:
       - id: ruff
         args: [--fix, --show-fixes]
+        exclude: "^tests/dummymodule.py"
 
   - repo: https://github.com/psf/black
     rev: 23.7.0
     hooks:
       - id: black
-        exclude: "^tests/mypy/negative.py"
+        exclude: "^tests/dummymodule.py|^tests/mypy/negative.py"
 
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v1.5.1
--- a/src/typeguard/_transformer.py
+++ b/src/typeguard/_transformer.py
@@ -57,6 +57,7 @@ from ast import (
     fix_missing_locations,
     iter_fields,
     keyword,
+    stmt,
     walk,
 )
 from collections import defaultdict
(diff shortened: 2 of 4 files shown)


agronholm:
Are the test suites passing for you locally?

vthemelis:
> Are the test suites passing for you locally?

Yes, I can get a clean run with `tox`. I'm just having trouble with linters being applied to places where they shouldn't by `pre-commit.ci` though which leads to redness on the PR.

vthemelis:
@agronholm All 💚 now!

agronholm:
> Yes, I can get a clean run with `tox`. I'm just having trouble with linters being applied to places where they shouldn't by `pre-commit.ci` though which leads to redness on the PR.

Installing pre-commit locally would have made that easier..

agronholm:
The test for this really belongs in `test_transformer.py`, not in `dummymodule.py`.

vthemelis:
> Installing pre-commit locally would have made that easier..

How do you run it locally? I tried `pre-commit run` but it was no-op for me.

vthemelis:
> The test for this really belongs in `test_transformer.py`, not in `dummymodule.py`.

It seemed like `dummymodule.py` gave me pretty good integration coverage already so I went this way. Do you want me to add unit tests explicitly to `test_transformer.py` too?

agronholm:
First of all, you should do `pre-commit install` which ensures that the checks are run when you do `git commit`. If you want to run the checks without the git hook, do `pre-commit run -a` (`-a` meaning "run for all files").

agronholm:
The problem seems to stem from `__post_init__()` only setting the right injection index if `self.node` is set, which it isn't when a module is being transformed by the import hook.

agronholm:
I'm likely to reject this PR in favor of an alternate approach and an alternate test.

agronholm:
Here's a test that fails without any patch:
```python
def test_respect_future_import() -> None:
    # Regression test for #385
    node = parse(
        dedent(
            '''
            from __future__ import annotations
            
            def foo() -> int:
                return 1
            '''
        )
    )
    TypeguardTransformer().visit(node)
    assert (
        unparse(node)
        == dedent(
            '''
            from __future__ import annotations
            from typeguard import TypeCheckMemo
            from typeguard._functions import check_return_type

            def foo() -> int:
                memo = TypeCheckMemo(globals(), locals())
                return check_return_type('foo', 1, int, memo)
            '''
        ).strip()
    )
```

vthemelis:
> I'm likely to reject this PR in favor of an alternate approach and an alternate test.

Fair enough! Let me know if you want me to help.

vthemelis:
> Here's a test that fails without any patch:

Thanks for this! Seems to pass with my changes. Shall I commit it?

agronholm:
I'm still considering my options. If I figure out a better way, I'll just push that and close this PR. Either way I'll let you know.

vthemelis:
> I'm still considering my options. If I figure out a better way, I'll just push that and close this PR. Either way I'll let you know.

Sounds good. I added the test in case we merge this PR.

agronholm:
Alright, I resolved this via c11057ab9f9570718fe149b2bfa6eef1df6f26a2. Thanks for your initiative on the matter.

vthemelis:
Thanks! Would it be possible to release the latest changes?

agronholm:
> Thanks! Would it be possible to release the latest changes?

Done.

PR #388, opened 2023-08-27
Only fill-in empty bodies

vthemelis:
Rather than any empty list-valued AST field.

This is an improvement on https://github.com/agronholm/typeguard/pull/386 given https://github.com/python/cpython/issues/108521#issuecomment-1694597859

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**UNRELEASED**
+
+- Only fill-in empty `body` blocks that may have been left empty by a ``@typechecked`` optimization
+  (`#352 <https://github.com/agronholm/typeguard/issues/352>`_)
+
 **4.1.3** (2023-08-27)
 
 - Dropped Python 3.7 support
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,11 +3,6 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
-**UNRELEASED**
-
-- Only fill-in empty `body` blocks that may have been left empty by a ``@typechecked`` optimization
-  (`#352 <https://github.com/agronholm/typeguard/issues/352>`_)
-
 **4.1.3** (2023-08-27)
 
 - Dropped Python 3.7 support
--- a/src/typeguard/_transformer.py
+++ b/src/typeguard/_transformer.py
@@ -55,7 +55,6 @@ from ast import (
     copy_location,
     expr,
     fix_missing_locations,
-    iter_fields,
     keyword,
     walk,
 )
@@ -500,17 +499,20 @@ class TypeguardTransformer(NodeTransformer):
         self.target_lineno = target_lineno
 
     def generic_visit(self, node: AST) -> AST:
-        non_empty_list_fields = []
-        for field_name, val in iter_fields(node):
-            if isinstance(val, list) and len(val) > 0:
-                non_empty_list_fields.append(field_name)
+        has_non_empty_body_initially = bool(getattr(node, "body", None))
+        initial_type = type(node)
 
         node = super().generic_visit(node)
 
-        # Add `pass` to list fields that were optimised away
-        for field_name in non_empty_list_fields:
-            if not getattr(node, field_name, None):
-                setattr(node, field_name, [Pass()])
+        if (
+            type(node) is initial_type
+            and has_non_empty_body_initially
+            and hasattr(node, "body")
+            and not node.body
+        ):
+            # If we have still the same node type after transformation
+            # but we've optimised it's body away, we add a `pass` statement.
+            node.body = [Pass()]
 
         return node
 


src/typeguard/_transformer.py
```diff
@@ -500,17 +505,27 @@ def __init__(
...
     def generic_visit(self, node: AST) -> AST:
-        non_empty_list_fields = []
-        for field_name, val in iter_fields(node):
-            if isinstance(val, list) and len(val) > 0:
-                non_empty_list_fields.append(field_name)
+        class HasBody(Protocol):
+            body: list[stmt]
+
+        def has_body(node: AST) -> TypeGuard[HasBody]:
+            return hasattr(node, "body") and node.body is not None
```
agronholm, line 512: Why all this here? What value does it add?

vthemelis: I just avoid adding these checks in my if statements. I can remove it if you prefer.

agronholm: But your if statements would run these checks anyway? I just don't understand why this is somehow better than `if getattr(node, "body", None):`.

vthemelis: The `if` statement would run these checks anyway but I only have to write on check now. I will rewrite without.

vthemelis: Changed

src/typeguard/_transformer.py
```diff
@@ -500,17 +499,23 @@ def __init__(
...
-        # Add `pass` to list fields that were optimised away
-        for field_name in non_empty_list_fields:
-            if not getattr(node, field_name, None):
-                setattr(node, field_name, [Pass()])
+        new_type = type(node)
+        if (
+            initial_type == new_type
+            and has_non_empty_body_initially
+            and hasattr(node, "body")
+            and (not node.body or len(node.body) == 0)
```
agronholm, line 514:
```suggestion
            and getattr(node, "body", None)
```

src/typeguard/_transformer.py
```diff
@@ -500,17 +499,23 @@ def __init__(
         self.target_lineno = target_lineno
 
     def generic_visit(self, node: AST) -> AST:
-        non_empty_list_fields = []
-        for field_name, val in iter_fields(node):
-            if isinstance(val, list) and len(val) > 0:
-                non_empty_list_fields.append(field_name)
+        has_non_empty_body_initially = (
+            hasattr(node, "body") and node.body and len(node.body) > 0
+        )
```
agronholm, line 504:
```suggestion
        has_non_empty_body_initially = bool(getattr(node, "body", None))
```

src/typeguard/_transformer.py
```diff
@@ -500,17 +499,23 @@ def __init__(
...
 
         node = super().generic_visit(node)
 
-        # Add `pass` to list fields that were optimised away
-        for field_name in non_empty_list_fields:
-            if not getattr(node, field_name, None):
-                setattr(node, field_name, [Pass()])
+        new_type = type(node)
+        if (
+            initial_type == new_type
```
agronholm, line 511:
```suggestion
        if (
            type(node) is initial_type
```

agronholm (commented):
Some simplifications.

src/typeguard/_transformer.py
vthemelis: added

src/typeguard/_transformer.py
vthemelis: This one is a bit more difficult. If I remove `hasattr(node, "body")`, then `mypy` cannot infer that the `node` actually has a `body`.

src/typeguard/_transformer.py
vthemelis: added

src/typeguard/_transformer.py
vthemelis: If you prefer, I can accept this change at the expense of either:
* Changing `node.body = ...` in the body of the if statement to a `setattr` equivalent with an annotation to prohibit `ruff` to changing back to normal assignment;
* Or adding a type ignore to pacify `mypy`.

agronholm: Then how about:
```
and hasattr(node, "body")
and not node.body
```
The `len(node.body) == 0` check is irrelevant because if the body was a sequence of length 0, it would already evaluate as falsy in the previous check.

vthemelis: done

agronholm:
Finally, was there a case where your previous code would have caused problems? What I mean to say, all tests were passing before this already.

vthemelis:
> Finally, was there a case where your previous code would have caused problems? What I mean to say, all tests were passing before this already.

Not that I know of but I thought that it's a good idea to add this defensively.

docs/versionhistory.rst
```diff
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**UNRELEASED**
+
+- Only fill-in empty `body` blocks that may have been left empty by a ``@typechecked`` optimization
+  (`#352 <https://github.com/agronholm/typeguard/issues/352>`_)
+
```
agronholm, line 10: If this doesn't fix an existing bug and isn't visible to users, then a changelog entry isn't necessary.
```suggestion
```

agronholm:
Thanks.

PR #389, opened 2023-08-27
Don't apply ruff and black on intentionally non-compliant modules

vthemelis:


agronholm:
Why?

agronholm:
`dummymodule` should be compliant with ruff and black, just not with mypy.

vthemelis:
Ah, indeed. The reason I was getting all the diffs with ruff yesterday in https://github.com/agronholm/typeguard/pull/387 was because of `from __future__ import annotations`. Will close.

Issue #384, opened 2023-08-22
Possible 4.1.2 regression: `AttributeError: 'BinOp' object has no attribute 'left'`


schmidt-ai:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Typeguard version

4.1.2

### Python version

3.10.12

### What happened?

When running `pytest` on my codebase with the `typeguard` plugin enabled, I get the following stack trace:

```txt
venv/lib/python3.10/site-packages/_pytest/runner.py:341: in from_call
    result: Optional[TResult] = func()
venv/lib/python3.10/site-packages/_pytest/runner.py:372: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
venv/lib/python3.10/site-packages/_pytest/python.py:531: in collect
    self._inject_setup_module_fixture()
venv/lib/python3.10/site-packages/_pytest/python.py:545: in _inject_setup_module_fixture
    self.obj, ("setUpModule", "setup_module")
venv/lib/python3.10/site-packages/_pytest/python.py:310: in obj
    self._obj = obj = self._getobj()
venv/lib/python3.10/site-packages/_pytest/python.py:528: in _getobj
    return self._importtestmodule()
venv/lib/python3.10/site-packages/_pytest/python.py:617: in _importtestmodule
    mod = import_path(self.path, mode=importmode, root=self.config.rootpath)
venv/lib/python3.10/site-packages/_pytest/pathlib.py:565: in import_path
    importlib.import_module(module_name)
../../.pyenv/versions/3.10.11/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1050: in _gcd_import
    ???
<frozen importlib._bootstrap>:1027: in _find_and_load
    ???
<frozen importlib._bootstrap>:1006: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:688: in _load_unlocked
    ???
venv/lib/python3.10/site-packages/_pytest/assertion/rewrite.py:178: in exec_module
    exec(co, module.__dict__)
tests/datapipe_test.py:7: in <module>
    from src.datapipe import Mask, TMACoreCropDataset
<frozen importlib._bootstrap>:1027: in _find_and_load
    ???
<frozen importlib._bootstrap>:1006: in _find_and_load_unlocked
    ???
[...]

Issue #337, opened 2023-04-11
Cannot check types that are shadowed by argument names


jolaf:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Typeguard version

4.0.0rc2 master

### Python version

3.10.6

### What happened?

```
Traceback (most recent call last):
  File "/media/sf_d/WWPass/git/python-test/test.py", line 4, in <module>
    import_module('A')
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "/home/jolaf/git/typeguard/src/typeguard/_importhook.py", line 122, in exec_module
    super().exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/media/sf_d/WWPass/git/python-test/A.py", line 6, in <module>
    f(bool)
  File "/media/sf_d/WWPass/git/python-test/A.py", line 1, in f
    def f(type: type | None = None) -> None:
  File "/home/jolaf/git/typeguard/src/typeguard/_functions.py", line 135, in check_argument_types
    check_type_internal(value, annotation, memo)
  File "/home/jolaf/git/typeguard/src/typeguard/_checkers.py", line 756, in check_type_internal
    checker(value, origin_type, args, memo)
  File "/home/jolaf/git/typeguard/src/typeguard/_checkers.py", line 432, in check_uniontype
    raise TypeCheckError(f"did not match any element in the union:\n{formatted_errors}")
typeguard.TypeCheckError: argument "type" (class bool) did not match any element in the union:
  bool: is not an instance of bool
  NoneType: is not an instance of NoneType
```

### How can we reproduce the bug?

`test.py`:
```python3
from importlib import import_module
from typeguard import install_import_hook
with install_import_hook(('A',)):
    import_module('A')
```

`A.py`:
```python3
def f(type: type | None = None) -> None:
    pass

f(bool)
```

Run:
```shell
$ python3 test.py
```

[...]

PR #392, opened 2023-08-28
Handle missing branches in `BinOp`

vthemelis:
Fixes https://github.com/agronholm/typeguard/issues/384 .

The issue is due to the `left` or `right` attributes of `BinOp` in the case that the `AnnotationTransformer` evaluates any of them to be `None`.

The code that removes the attributes completely is:
https://github.com/python/cpython/blob/e407cea1938b80b1d469f148a4ea65587820e3eb/Lib/ast.py#L493-L498

This patch removes the crush but I think shows another bug where in the test I added, the argument `A` is expected to only have a value of `None`. NB that this is not just wrong but also inconsistent with what happens when using `Union` instead of `|`

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,9 +3,9 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
-**unreleased**
+**UNRELEASED**
 
-- Fixed handling of `BinOp` nodes with missing branches in AnnotationTransformer
+- Fixed ``AttributeError`` where the transformer removed elements from a PEP 604 union
   (`#384 <https://github.com/agronholm/typeguard/issues/384>`_)
 
 **4.1.3** (2023-08-27)
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**unreleased**
+
+- Fixed handling of `BinOp` nodes with missing branches in AnnotationTransformer
+  (`#384 <https://github.com/agronholm/typeguard/issues/384>`_)
+
 **4.1.3** (2023-08-27)
 
 - Dropped Python 3.7 support
--- a/src/typeguard/_transformer.py
+++ b/src/typeguard/_transformer.py
@@ -380,9 +380,9 @@ class AnnotationTransformer(NodeTransformer):
             # If this happens, treat the BinOp as just the other branch.
             if not hasattr(node, "left") and not hasattr(node, "right"):
                 return None
-            if not hasattr(node, "left"):
+            elif not hasattr(node, "left"):
                 return node.right
-            if not hasattr(node, "right"):
+            elif not hasattr(node, "right"):
                 return node.left
 
             # Return Any if either side is Any
--- a/src/typeguard/_transformer.py
+++ b/src/typeguard/_transformer.py
@@ -376,6 +376,16 @@ class AnnotationTransformer(NodeTransformer):
         self.generic_visit(node)
 
         if isinstance(node.op, BitOr):
+            # If either branch of the BinOp has been transformed to `None`
+            # then the `ast.generic_visit` will eliminate that branch completely.
+            # If this happens, treat the BinOp as just the other branch.
+            if not hasattr(node, "left") and not hasattr(node, "right"):
+                return None
+            if not hasattr(node, "left"):
+                return node.right
+            if not hasattr(node, "right"):
+                return node.left
+
             # Return Any if either side is Any
             if self._memo.name_matches(node.left, *anytype_names):
                 return node.left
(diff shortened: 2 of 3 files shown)


tests/test_transformer.py
```diff
@@ -1570,6 +1570,68 @@ def foo(x: str) -> None:
...
+        unparse(node)
+        == dedent(
+            """
+            from __future__ import annotations
+
+            def foo(A: A | None) -> None:
+                from typeguard import TypeCheckMemo
+                from typeguard._functions import check_argument_types
+                memo = TypeCheckMemo(globals(), locals())
+                check_argument_types('foo', {'A': (A, None)}, memo)
```
vthemelis, line 1629: This doesn't look right and it's inconsistent with what we get when using `Union` like in the previous test.

Perhaps I should change the code to return `None` if either of the `BinOp` branches has been erased for now?

agronholm: I'll take a good look at this tomorrow. Some notes:

- The first test passes without your patch
- Did you intentionally name the class and the argument the same? I'm not sure how that makes any sense.
- What, in your opinion, should the result look like?

vthemelis: Hi, thanks for the quick review!
> The first test passes without your patch
Yes, I added the first test to compare the output of `Union` and `|` to see if they are consistent. It can be removed but I think that it helps with the conversation.

> Did you intentionally name the class and the argument the same? I'm not sure how that makes any sense.
Yes, I did that intentionally as the easiest way to get a `None` transformation of either branch of the `BinOp`.

> What, in your opinion, should the result look like?
It should look the same as the one for `Union` I think. I do understand that there are some issues with shadowed types though: https://github.com/agronholm/typeguard/issues/337

vthemelis: Btw, I haven't tried it yet but maybe one way to get access to the shadowed type inside the function is to do something like:
```
from typing import Type

Type[A].__origin__
```

vthemelis: Interesting:
```Python
@typechecked
def foo(int: Union[int, bool]) -> int:
    return int

foo(42)
```
gives
```
TypeError: Union[arg, ...]: each arg must be a type. Got 42.
```

So I guess that the behaviour with `Union` is also incorrect due to the shadowing of the type.

agronholm: How about this: cut out the first test, leave the second in. This PR does demonstrably fix a bug, so let's just get this in. If you'd just add a note in the changelog, I'll merge. Any other union related behaviors should be addressed in a separate issue/PR.

vthemelis: I don't think that this PR 'fixes' a bug as it is; it only replaces one bug for another one.

In the example above, only `None` is an accepted value for `foo` whereas probably only value should be accepted (as we should effectively handle `int` as `Any` in this case.

This is demonstrated on this test:

https://github.com/agronholm/typeguard/pull/392/commits/5b5f91f3c5926fd2697b8381813c187e3c526548#diff-00ca1530aa0e4a5f0915df54f28fd180973f39b478af9b8beca7b49884903cd5R675-R685

(currently failing)

agronholm: I think that needs a separate bugfix. This PR could be merged to just fix the transformer issue.

vthemelis: @agronholm, I removed the failing test and the classic union test. I'll add an issue about the `|` bug after we merge this.

docs/versionhistory.rst
```diff
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**unreleased**
```
agronholm, line 6:
```suggestion
**UNRELEASED**
```

docs/versionhistory.rst
```diff
@@ -3,6 +3,11 @@ Version history
...
+
+- Fixed handling of `BinOp` nodes with missing branches in AnnotationTransformer
```
agronholm, line 8:
```suggestion
- Fixed ``AttributeError`` where the transformer removed elements from a PEP 604 union
```

src/typeguard/_transformer.py
```diff
@@ -375,6 +375,16 @@ def visit_BinOp(self, node: BinOp) -> Any:
...
 
         if isinstance(node.op, BitOr):
+            # If either branch of the BinOp has been transformed to `None`
+            # then the `ast.generic_visit` will eliminate that branch completely.
+            # If this happens, treat the BinOp as just the other branch.
+            if not hasattr(node, "left") and not hasattr(node, "right"):
+                return None
+            if not hasattr(node, "left"):
+                return node.right
+            if not hasattr(node, "right"):
```
agronholm, line 385: This is just for aesthetics, but I prefer separate control blocks to be separated by a blank line.
```suggestion
            elif not hasattr(node, "left"):
                return node.right
            elif not hasattr(node, "right"):
```

agronholm:
Thanks!

tests/test_transformer.py
vthemelis: Added https://github.com/agronholm/typeguard/issues/395

PR #408, opened 2023-09-23
Avoid creating ref cycles

hauntsaninja:
By storing previously raised exceptions inside a local, this code created ref cycles that kept all locals in all calling stack frames alive.

This is because exceptions hold references to their tracebacks, which hold references to the relevant frames, which holds a reference to the local errors dict that holds references to the exceptions. See https://peps.python.org/pep-0344/#open-issue-garbage-collection and https://peps.python.org/pep-3110/#rationale

This breaks the cycle by deleting the local when we raise, so frames are destroyed by the normal reference counting mechanism.

This fixes some resource exhaustion issues I encountered at work.

### DIFF
--- a/src/typeguard/_checkers.py
+++ b/src/typeguard/_checkers.py
@@ -405,7 +405,10 @@ def check_union(
     formatted_errors = indent(
         "\n".join(f"{key}: {error}" for key, error in errors.items()), "  "
     )
-    raise TypeCheckError(f"did not match any element in the union:\n{formatted_errors}")
+    try:
+        raise TypeCheckError(f"did not match any element in the union:\n{formatted_errors}")
+    finally:
+        del errors
 
 
 def check_uniontype(


agronholm:
A things of note:

1. Why not delete the errors before returning?
2. I would like to see a test that ensures there won't be reference cycles

hauntsaninja:
Thanks for the review! I must have missed your response. I've addressed both points now :-)

agronholm:
Would you mind also adding a note in the changelog (under `**UNRELEASED**`)? Then we're good.

hauntsaninja (commented):
Thanks, added the changelog, a fix, and a test for the fix

docs/versionhistory.rst
```diff
@@ -4,6 +4,11 @@ Version history
 This library adheres to
 `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**UNRELEASED**
+
+- Avoid creating reference cycles when type checking unions
+  (`#408 <https://github.com/agronholm/typeguard/pull/408>`_)
```
agronholm, line 10: If there was no issue being resolved by this, then the link should go away.

hauntsaninja: Sorry about that, removed!

agronholm:
Thanks!

PR #433, opened 2024-01-24
Fixed generic aliases not passing checks against `type`

### LINKED ISSUE #432: False positive when checking `<class 'types.GenericAlias'>`(e.g.`dict[str, str]`) against `type`
### Things to check first

### Typeguard version

4.1.5

### Python version

3.12.0

### What happened?

With this file:
```python
import typeguard

@typeguard.typechecked
def foo(t: type):
    pass

foo(dict[str,str])
```

typeguard reports:

```
  File "/****/test.py", line 7, in <module>
    foo(dict[str,str])
  File "/****/test.py", line 4, in foo
    def foo(t: type):
  File "/****/lib/python3.12/site-packages/typeguard/_functions.py", line 138, in check_argument_types
    check_type_internal(value, annotation, memo)
  File "/****/lib/python3.12/site-packages/typeguard/_checkers.py", line 759, in check_type_internal
  File "/****/lib/python3.12/site-packages/typeguard/_checkers.py", line 438, in check_class
    raise TypeCheckError("is not a class")
typeguard.TypeCheckError: argument "t" (types.GenericAlias) is not a class
```

It should pass the test

With this file:
```python
import typeguard

@typeguard.typechecked
def foo(t: type):
    pass

foo(dict[str,str])
```

typeguard reports:

```
  File "/****/test.py", line 7, in <module>
    foo(dict[str,str])
  File "/****/test.py", line 4, in foo
    def foo(t: type):
  File "/****/lib/python3.12/site-packages/typeguard/_functions.py", line 138, in check_argument_types
    check_type_internal(value, annotation, memo)
  File "/****/lib/python3.12/site-packages/typeguard/_checkers.py", line 759, in check_type_internal
  File "/****/lib/python3.12/site-packages/typeguard/_checkers.py", line 438, in check_class
    raise TypeCheckError("is not a class")
typeguard.TypeCheckError: argument "t" (types.GenericAlias) is not a class
```

fefe982:
fixes #432

`is_class(dict[str,str], type)` returns `False` after python 3.11. Use `isinstance(dict[str, str], types.GenericAlias)` to check in this situation.

### DIFF
--- a/tests/test_checkers.py
+++ b/tests/test_checkers.py
@@ -889,6 +889,10 @@ class TestType:
     def test_union_typevar(self):
         T = TypeVar("T", bound=Parent)
         check_type(Child, Type[T])
+    
+    def test_generic_aliase(self):
+        if sys.version_info >= (3, 9):
+            check_type(dict[str, str], type)
 
 
 class TestIO:
--- a/tests/test_checkers.py
+++ b/tests/test_checkers.py
@@ -889,7 +889,7 @@ class TestType:
     def test_union_typevar(self):
         T = TypeVar("T", bound=Parent)
         check_type(Child, Type[T])
-    
+
     def test_generic_aliase(self):
         if sys.version_info >= (3, 9):
             check_type(dict[str, str], type)
--- a/tests/test_checkers.py
+++ b/tests/test_checkers.py
@@ -893,6 +893,8 @@ class TestType:
     def test_generic_aliase(self):
         if sys.version_info >= (3, 9):
             check_type(dict[str, str], type)
+        check_type(Dict, Type[Any])
+        check_type(Dict[str, str], Type[Any])
 
 
 class TestIO:
--- a/tests/test_checkers.py
+++ b/tests/test_checkers.py
@@ -891,11 +891,13 @@ class TestType:
         T = TypeVar("T", bound=Parent)
         check_type(Child, Type[T])
 
-    def test_generic_aliase(self):
+    @pytest.mark.parametrize("check_against", [type, Type[Any]])
+    def test_generic_aliase(self, check_against):
         if sys.version_info >= (3, 9):
-            check_type(dict[str, str], type)
-        check_type(Dict, Type[Any])
-        check_type(Dict[str, str], Type[Any])
+            check_type(dict[str, str], check_against)
+
+        check_type(Dict, check_against)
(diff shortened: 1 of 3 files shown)


tests/test_checkers.py
```diff
@@ -890,6 +890,10 @@ def test_union_typevar(self):
         T = TypeVar("T", bound=Parent)
         check_type(Child, Type[T])
 
+    def test_generic_aliase(self):
+        if sys.version_info >= (3, 9):
+            check_type(dict[str, str], type)
```
agronholm, line 895: Wouldn't this be better?
```suggestion
    def test_generic_aliases(self):
        check_type(Dict[str, str], Type[Any])
```

fefe982: This check raises another problem: `Dict[str, str]` is of type `typing._GenericAlias`, not `types.GenericAlias`. It cannot be covered by this PR.

I'm not using generic types from typing, and didn't notice this problem.

fefe982: The typing module comes in python 3.5. The github repo for cpython has tags from 3.8.

From there, I see that in 3.8, `Dict` and `Dict[str, str]` are all of type `typing._GenericAlias`. From 3.9, `Dict` has type `typing._SpecialGenericAlias` and `Dict[str, str]` is of type `typing._GenerecAlias`, both are inherited from `typing._BaseGenericAlias`. `_BaseGenericAlias` does not exist in python 3.8. And as the name shows, they are not part of the public interface of `typing`. I think I can only check for these types with:

```python
isinstance(Dict[str, str], type(Type[Any])) # typing._GenericAlias
isinstance(Dict, type(Type)) # typing._SpecialGenericAlias
```

I think all of them should be checked and tested.

Also, this might not work for older python version. I found in python 3.6, `Dict` inherited from `dict` (and some other classes), and `_GenericAlias` does not exit. I did not find a python 3.5.

I only run the actual test in python 3.11.

agronholm: Any Python version older than 3.8 is irrelevant here. And even if those private types from `typing` aren't public, they can still be used as long as we confine that use to below a maximum Python version (3.8?).

agronholm: I've tweaked the checks and the tests somewhat. I think it's ripe for merging now.

agronholm:
Thanks!

PR #460, opened 2024-05-25
Support Python 3.13

JelleZijlstra:
- typing.is_typeddict() does not recognize typing_extensions.TypedDict
- ForwardRef._evaluate gained another parameter

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,6 +6,8 @@ This library adheres to
 
 **UNRELEASED**
 
+- Fixed some compatibility problems when running on Python 3.13
+  (`#460 <https://github.com/agronholm/typeguard/issues/460>`_)
 - Fixed test suite incompatibility with pytest 8.2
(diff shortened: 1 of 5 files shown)


JelleZijlstra:
The test failures are due to some pytest deprecation; it's not immediately obvious where the problem is.

JelleZijlstra:
The good news is that my draft PEP 649 implementation (https://github.com/python/cpython/pull/119361) passes the entire test suite, once this change is implied and I ignore the pytest deprecation.

agronholm:
This needs additional tests. If `typing_extensions` is to be imported unconditionally, then the conditional dependency on it in `pyproject.toml` must be made unconditional.

JelleZijlstra:
The test suite already tests for both `typing.TypedDict` and `typing_extensions.TypedDict`; it uses the `typing_provider` fixture to use both. However, I'll add a direct test for `is_typeddict`.

agronholm:
Thanks!!

PR #464, opened 2024-05-28
Fix `pytest_ignore_collect` not to block default pytest code

mgorny:
## Changes

Fix `pytest_ignore_collect` hook implementation to return `None` rather than `False` when the path ought not to be ignored.  This is necessary to enable the default pytest implementation of `pytest_ignore_collect()` to be used.  Otherwise, the default rules are never applied and e.g. `--ignore` does not work.

I've also reported https://github.com/pytest-dev/pytest/issues/12383 about the misleading documentation.

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -22,6 +22,9 @@ This library adheres to
 - Fixed checks against annotations wrapped in ``NotRequired`` not being run unless the
   ``NotRequired`` is a forward reference
   (`#454 <https://github.com/agronholm/typeguard/issues/454>`_)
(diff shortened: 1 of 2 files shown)


agronholm:
Thanks! But next time, please add a changelog entry yourself so I won't have to.

mgorny:
Ah, sorry. I guess I thought that it's not a "user-facing change", since it's just the test suite.

Thanks!

agronholm:
> Ah, sorry. I guess I thought that it's not a "user-facing change", since it's just the test suite.
> 
> Thanks!

Sorry, this was my mistake. The changelog entry I added is invalid.

PR #470, opened 2024-06-28
Add extra tests for __future__ annotations

epenet:
Linked to https://github.com/hacf-fr/renault-api/pull/1234

## Changes

Fixes #. 

## Checklist

features)
- [ ] You've added a new changelog entry (in `docs/versionhistory.rst`).

these instructions.

### Updating the changelog

### DIFF
--- a/tests/test_transformer.py
+++ b/tests/test_transformer.py
@@ -1854,7 +1854,6 @@ def test_respect_future_import_carriage_return() -> None:
         == dedent(
             '''
             """module docstring"""
-
             from __future__ import annotations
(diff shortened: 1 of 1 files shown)


tests/test_transformer.py
```diff
@@ -1834,6 +1834,39 @@ def foo() -> int:
...
+            '''
+        )
+    )
+    TypeguardTransformer().visit(node)
+    assert (
+        unparse(node)
+        == dedent(
+            '''
+            """module docstring"""
+
```
epenet, line 1857:
```suggestion
```

Issue #472, opened 2024-07-10
typing_extensions.Buffer requires `typing-extensions>=4.6`


ymerkli:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Typeguard version

4.2.1

### Python version

3.10.12

### What happened?

typeguard uses `typing_extensions.Buffer`, which was only added to `typing_extenions>=4.6.0`.

### How can we reproduce the bug?

```
pip install typing_extensions==4.5.0
python -c "from typeguard import check_type"
```

agronholm:
Typeguard already depends on `typing_extensions >= 4.10.0`, so I have no clue why you would report this as a bug.

PR #473, opened 2024-07-10
fix: lower bound typing-extensions >= 4.6

ymerkli:
## Changes

Fixes https://github.com/agronholm/typeguard/issues/472

typeguard uses `typing_extensions.Buffer`, which was only added to `typing_extenions>=4.6.0`, [see here](https://github.com/python/typing_extensions/blob/main/CHANGELOG.md#release-460-may-22-2023).

This PR adds `typing_extenions>=4.6.0` as a requirement.

## Checklist

- [x] If this is a trivial change, like a typo fix or a code reformatting, then you can ignore
these instructions.
- [x] Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -4,6 +4,11 @@ Version history
 This library adheres to
 `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**UNRELEASED**
+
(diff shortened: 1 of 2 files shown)


agronholm:
Typeguard already has a higher `typing_extensions` requirement [here](https://github.com/agronholm/typeguard/blob/master/pyproject.toml#L29). Besides, your PR adds it to build requirements, why?

PR #475, opened 2024-07-31
Use environment variable TYPEGUARD_DISABLE to disable typeguard.

tage64:
## Changes

Setting the environment variable `TYPEGUARD_DISABLE` disables Typeguard and has the same effect as running Python with the `-O` flag.

I wanted to compile my Python program with [Nuitka](https://nuitka.net) which removes the source code so that the `inspect` module doesn't work. However, I did not want to disable all assertians with the `-O` flag.

## Checklist

- [X] You've added tests (in `tests/`) added which would fail without your patch
- [X] You've updated the documentation (in `docs/`, in case of behavior changes or new
features)
- [X] You've added a new changelog entry (in `docs/versionhistory.rst`).

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -261,8 +261,8 @@ Suppressing the ``@typechecked`` decorator in production
 If you're using the :func:`@typechecked <typechecked>` decorator to gradually introduce
 run-time type checks to your code base, you can disable the checks in production by
 running Python in optimized mode (as opposed to debug mode which is the default mode).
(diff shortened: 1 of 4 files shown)


agronholm:
Because 1) I have plenty of other projects to maintain too, 2) I want to understand the feature and decide if I want it and if the it's the right approach.

JohannesK71083:
Ok, I just wanted to know if there is any work to do to make this PR acceptable. Apparently not at this moment, thanks for the reply.

PR #478, opened 2024-08-26
Support classmethod w/ pos-only arguments

jermenkoo:
## Changes

While upgrading `typeguard` internally at our company, I noticed it does not handle `classmethod`s with positional-only arguments. While I have no idea why someone would need that, I could not find online why this should not be done. This PR fixes this issue.

## Checklist

features)

these instructions.

### Updating the changelog

Will update changelog once we agree on the approach here.

### DIFF
--- a/src/typeguard/_transformer.py
+++ b/src/typeguard/_transformer.py
@@ -863,8 +863,9 @@ class TypeguardTransformer(NodeTransformer):
                             isinstance(decorator, Name)
                             and decorator.id == "classmethod"
                         ):
+                            arglist = node.args.args or node.args.posonlyargs
                             memo_kwargs["self_type"] = Name(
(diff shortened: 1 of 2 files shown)


agronholm:
You checked the box for having made documentation changes and added a changelog entry, but I'm not seeing any? Why did you check those boxes if you didn't do that?

agronholm:
I'm going to reject this in favor of a more comprehensive changeset which I'll be making. It will add this fix for instance methods too (no point in only supporting class methods, yes?).

On another note, the line `arglist = node.args.args or node.args.posonlyargs` was wrong (the operands should have been reversed, as `self` is always a positional-only argument if there are any of them. The accompanying test was not comprehensive enough to pick up on this.

PR #485, opened 2024-09-08
Fixed annotation for `typeguard_ignore()` to match one for `typing.no_type_check()`

jolaf:
Trivial fix to make sure `typeguard_ignore()` has the same type signature in both `TYPE_CHECKING` and runtime branches.

### DIFF
--- a/src/typeguard/_decorators.py
+++ b/src/typeguard/_decorators.py
@@ -16,19 +16,17 @@ from ._functions import TypeCheckFailCallback
 from ._transformer import TypeguardTransformer
 from ._utils import Unset, function_name, get_stacklevel, is_method_of, unset
 
+T_CallableOrType = TypeVar("T_CallableOrType", bound=Callable[..., Any])
+
 if TYPE_CHECKING:
     from typeshed.stdlib.types import _Cell
 
-    _F = TypeVar("_F")
-
-    def typeguard_ignore(f: _F) -> _F:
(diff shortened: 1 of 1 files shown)


jolaf:
Here's the test, demonstrating the issue.

`test.py`:
```python3
with install_import_hook(('A',)):
    import A
```
`A.py`:
```python3
try:
    from typeguard import typeguard_ignore
except ImportError:
    from typing import no_type_check as typeguard_ignore

@typeguard_ignore
def f() -> None:
    a: int = "OK"
    print(a)

f()
```

```shell
$ python3 A.py
OK

$ python3 test.py
OK

$ mypy A.py
A.py:4:5: error: Incompatible import of "typeguard_ignore" (imported name has type "Callable[[_F@no_type_check], _F@no_type_check]", local name has type "Callable[[_F@typeguard_ignore], _F@typeguard_ignore]")  [assignment]
        from typing import no_type_check as typeguard_ignore
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
A.py: note: In function "f":
A.py:8:14: error: Incompatible types in assignment (expression has type "str", variable has type "int")  [assignment]
        a: int = "OK"
                 ^~~~
Found 2 errors in 1 file (checked 1 source file)
```
The second error is expected, but the first one is not. It goes away with this patch.

agronholm:
Thanks!

PR #488, opened 2024-09-20
Fix tests to work on Windows.

thetorpedodog:
## Changes

A few tests fail on a clean checkout on Windows due to formatting differences (CRLF newlines and \ in filenames). These should now pass on both Windows and Unix/Linux platforms.

## Checklist

- [n/a] You've added tests (in `tests/`) added which would fail without your patch
  (these are the tests that are already there!)
- [n/a] You've updated the documentation (in `docs/`, in case of behavior changes or new features)
  (no behavior changes or new features)
- [n/a] You've added a new changelog entry (in `docs/versionhistory.rst`).
  (trivial)


agronholm:
Nice, thanks!

Issue #465, opened 2024-05-29
Incorrect checking of protocol methods with `**kwargs` but no `*args`


comex:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Typeguard version

4.3.0

### Python version

3.12

### What happened?

The below example should pass checking, but instead gives:
```
Traceback (most recent call last):
  File "/Users/comex/src/typeguard/test.py", line 12, in <module>
    check_type(C(), P)
  File "/Users/comex/src/typeguard/src/typeguard/_functions.py", line 106, in check_type
    check_type_internal(value, expected_type, memo)
  File "/Users/comex/src/typeguard/src/typeguard/_checkers.py", line 861, in check_type_internal
    checker(value, origin_type, args, memo)
  File "/Users/comex/src/typeguard/src/typeguard/_checkers.py", line 729, in check_protocol
    raise TypeCheckError(
typeguard.TypeCheckError: __main__.C is not compatible with the P protocol because its 'foo' method has too few arguments in its declaration; expected 2 but 1 argument(s) declared
```

### How can we reproduce the bug?

```python
from typeguard import check_type
from typing import Protocol

class P(Protocol):
    def foo(self, **kwargs):
        ...

class C:
    def foo(self, **kwargs):
        pass

check_type(C(), P)
```

agronholm:
I suspect that this isn't the only issue with the new protocol check. I'll expand the test suite for that a bit and fix any problems I find in the process.

jace:
The new protocol check implementation also doesn't work with callback protocols because it checks against the type. This fails with `typeguard.TypeCheckError: function is not compatible with the MyProtocol protocol because it has no attribute named '__weakref__'`

agronholm:
> The new protocol check implementation also doesn't work with callback protocols because it checks against the type. This fails with `typeguard.TypeCheckError: function is not compatible with the MyProtocol protocol because it has no attribute named '__weakref__'`

Would you mind writing an example for me so I could make sure it works in the next release?

jkulhanek:
Having the same issue.

agronholm:
> Having the same issue.

Then you can just subscribe to the issue instead of pinging everyone.

jkulhanek:
> > Having the same issue.
> 
> Then you can just subscribe to the issue instead of pinging everyone.

So sorry, I will keep it in mind for next time.

[...]

PR #490, opened 2024-09-21
Quickfix for intersection protocols

antonagestam:
## Changes

This applies a quickfix solution for interesection protocols, opting to fully ignore runtime checks of `__orig_bases__` and `__weakref__` attributes. This has the effect of turning some false positives into true negatives, but it also leaves some false negatives. To make that clear, xfail test cases are added for the resulting false negatives.

A more robust fix would perhaps be to introduce recursive behavior whenever a `__orig_bases__` attribute is found, making the type checker check each inherited protocol individually. The resulting "top level" intersection type would also need individual checking though, as it can add required attributes on its own.

## Checklist

features)

these instructions.

### Updating the changelog


agronholm:
Shouldn't the proper fix involve `typing_extensions.get_protocol_members()`?

antonagestam:
> Shouldn't the proper fix involve `typing_extensions.get_protocol_members()`?

Ah I see, thanks. I was not aware of that utility. Would something like this change make sense if it used get_protocol_members? I mean specifically about the xfails introduced. Or would you rather also see those addressed all in one go?

agronholm:
Would it require a lot more effort to fix all of it?

antonagestam:
@agronholm I think it might be a larger effort, and I'm now not sure it's the correct thing to do either, perhaps protocol checking should really be "flat". Making those test cases pass seems more related to supporting type variables.

I think this PR is reasonably scoped now.

agronholm:
I'm currently running my own, uncommitted protocol tests against this branch. I'm hoping it will fix other protocol issues too.

tests/test_checkers.py
```diff
@@ -995,6 +998,59 @@ def test_text_real_file(self, tmp_path: Path):
...
+    SIT = TypeVar("SIT", bound=object, covariant=True)
+
+    class SizedIterable(
+        Sized,
+        Iterable[SIT],
+        Protocol[SIT],
+    ): ...
+
+    @pytest.mark.parametrize(
+        ("subject", "predicate_type"),
```
agronholm, line 1011: Follow the style of existing parametrization calls.
```suggestion
        "subject, predicate_type",
```

tests/test_checkers.py
```diff
@@ -995,6 +998,59 @@ def test_text_real_file(self, tmp_path: Path):
...
+        (
+            ((), SizedIterable),
+            (range(2), SizedIterable),
+            ((), SizedIterable[int]),
+            ((1, 2, 3), SizedIterable[int]),
+            (("1", "2", "3"), SizedIterable[str]),
```
agronholm, line 1017: Add `pytest.param() and the `id=` parameter to distinguish between individual test cases.

tests/test_checkers.py
```diff
@@ -995,6 +998,59 @@ def test_text_real_file(self, tmp_path: Path):
...
+                (1, 2, 3),
+                SizedIterable[str],
+                marks=xfail_nested_protocol_checks,
+            ),
+            pytest.param(
+                ("1", "2", "3"),
+                SizedIterable[int],
+                marks=xfail_nested_protocol_checks,
+            ),
+        ),
```
agronholm, line 1047: As above.

tests/test_checkers.py
```diff
@@ -995,6 +998,59 @@ def test_text_real_file(self, tmp_path: Path):
             check_type(f, TextIO)
 
 
+class TestIntersectingProtocol:
+    SIT = TypeVar("SIT", bound=object, covariant=True)
```
agronholm, line 1002: Does `bound=object` really do anything?

agronholm (changes requested):
Needs some code style updates.

agronholm:
Once you've addressed my comments, I'll merge this and add further fixes to Protocol support.

agronholm (approved):

antonagestam:
@agronholm Thanks for reviewing, the comments are now addressed 👍

agronholm:
Thanks!

ryan-williams:
@agronholm sorry to bug, just wondering if there's any plans for a release that will include this change?

agronholm:
I still need to fix https://github.com/agronholm/typeguard/issues/465 as it's a regression from the previous release.

agronholm:
If you want to speed things up, you could help review the [new PR](https://github.com/agronholm/typeguard/pull/496).

PR #491, opened 2024-09-22
Add tests for empty protocols on built-in objects.

thetorpedodog:
One of the symptoms of the earlier problem was that you could not check protocols against built-in types at all. This adds a test specifically for that situation to prevent a regression.

## Checklist

- [n/a] You've added tests (in `tests/`) added which would fail without your patch
  Regression test for change #490.
- [n/a] You've updated the documentation (in `docs/`, in case of behavior changes or new features)
- [n/a] You've added a new changelog entry (in `docs/versionhistory.rst`).

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -57,3 +57,12 @@ class RuntimeProtocol(Protocol):
     member: int
 
     def meth(self, x: str) -> None: ...
+
+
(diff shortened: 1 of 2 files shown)


agronholm:
I've added tests against an empty protocol, but I had to make such significant changes that I made a totally separate changeset.

PR #493, opened 2024-10-04
Fix ref cycles

ojw28:
This change is essentially the same fix as in https://github.com/agronholm/typeguard/pull/408,
applied in two additional functions that use the same pattern.


hauntsaninja (approved):

agronholm:
Thanks!

PR #496, opened 2024-10-26
Added proper Protocol method signature checking

### LINKED ISSUE #465: Incorrect checking of protocol methods with `**kwargs` but no `*args`

agronholm:
## Changes

Fixes #465.

This adds proper checking of protocol method signatures instead of using `check_callable()` which does not deal with keyword arguments.

## Checklist

- [X] You've added tests (in `tests/`) added which would fail without your patch
- [X] You've updated the documentation (in `docs/`, in case of behavior changes or new
features)
- [X] You've added a new changelog entry (in `docs/versionhistory.rst`).

these instructions.

### Updating the changelog

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -65,13 +65,8 @@ As of version 4.3.0, Typeguard can check instances and classes against Protocols
 regardless of whether they were annotated with
 :func:`@runtime_checkable <typing.runtime_checkable>`.
 
-There are several limitations on the checks performed, however:
-
-* For non-callable members, only presence is checked for; no type compatibility checks
-  are performed
-* For methods, only the number of positional arguments are checked against, so any added
-  keyword-only arguments without defaults don't currently trip the checker
-* Likewise, argument types are not checked for compatibility
+The only current limitation is that argument annotations are not checked for
+compatibility, however this should be covered by static type checkers pretty well.
 
 Special considerations for ``if TYPE_CHECKING:``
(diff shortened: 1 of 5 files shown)


antonagestam:
I can run the phantom-types test suites against this PR, will try and report back today.

antonagestam:
phantom-types test suite is passing without issues ✅

agronholm:
Thanks!

PR #497, opened 2024-10-29
Further improvement on typeguard_ignore() annotation

jolaf:
Unfortunately, it turned out that #485 was not enough to make `mypy` fully happy.
So this is a followup.

See https://github.com/python/mypy/issues/18060 for details.


agronholm:
The previous PR was missing the changelog note. I think you should add one, at least now that the problem has been properly fixed.

jolaf:
Which file should I add it to?

agronholm:
Was the pull request template not visible? The file is `docs/versionhistory.rst`.

PR #507, opened 2025-01-15
use full name for Attributes

### LINKED ISSUE #506: TypeCheckError when tuple unpacking to properties of method parameter
### Things to check first

### Typeguard version

4.4.1

### Python version

3.12.2

### What happened?

When using Typeguard 4.4.1 (does not happen in 4.4.0), tuple unpacking to attributes of an object that was passed as a parameter raises an incorrect TypeCheckError:

`typeguard.TypeCheckError: value assigned to <param>. <attr> (<type>) is not an instance of <class>`

The error message incorrectly claims we're trying to assign a value to the entire object, when we're actually assigning to one of its attributes.

The same operation works correctly:
- When using individual attribute assignments (`<param>.<attr> = <value>`)
- When tuple unpacking to attributes of locally created objects
- In Typeguard version 4.4.0

```python
"""Bug in Typeguard 4.4.1. Does not happen in 4.4.0."""

from dataclasses import dataclass


@dataclass
class Foo:
    bar: str = "bar"
    baz: int = 123


@typechecked
def run() -> None:
    foo = Foo()

    # Case 1: Direct attribute assignment works fine.
    foo.bar = "bar"
    foo.baz = 123

    # Case 2: Tuple unpacking in main scope works fine.
    foo.bar, foo.baz = ("bar", 123)

    mutate(foo)


@typechecked
def mutate(obj: Foo) -> None:
    # Case 3: Direct attribute assignment in method works fine.
    obj.bar = "bar"
    obj.baz = 123

    # Case 4: BUG - Tuple unpacking in method raises incorrect error.
    # Expected: Should work like Case 2.
    # Actual: Raises typeguard.TypeCheckError claiming str is not an instance of Foo.
    obj.bar, obj.baz = ("bar", 123)


if __name__ == "__main__":
    run()
```

fefe982:
## Changes

Fixes #506

In assignment check, when attribute is added:

1. `Attribute` is only handled in tuple, but not bare attribute assignments. Add check for bare attribute assignment
2. `exp.id` is used for `Attribute`, but for `Attribute`, it points to the parent object, not the attribute. Changed it to use `name` ，which is `obj.attr`, the full path.

## Checklist

features)

### DIFF
--- a/src/typeguard/_functions.py
+++ b/src/typeguard/_functions.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
+import inspect
 import sys
 import warnings
 from collections.abc import Sequence
 from typing import Any, Callable, NoReturn, TypeVar, Union, overload
-import inspect
 
 from . import _suppression
 from ._checkers import BINARY_MAGIC_METHODS, check_type_internal
--- a/src/typeguard/_functions.py
+++ b/src/typeguard/_functions.py
@@ -4,6 +4,7 @@ import sys
 import warnings
 from collections.abc import Sequence
 from typing import Any, Callable, NoReturn, TypeVar, Union, overload
+import inspect
 
 from . import _suppression
 from ._checkers import BINARY_MAGIC_METHODS, check_type_internal
@@ -280,7 +281,20 @@ def check_variable_assignment(
 
         for val, (varname, annotation) in zip(values_to_check, target):
             try:
-                check_type_internal(val, annotation, memo)
+                varpath = varname.split(".")
+                if len(varpath) > 1 and not inspect.isclass(annotation):
+                    continue
+                have_annotation_for_attribute = True
+                for seg in varpath[1:]:
+                    if not inspect.isclass(annotation):
+                        have_annotation_for_attribute = False
+                        break
+                    if sys.version_info >= (3, 10):
+                        annotation = inspect.get_annotations(annotation).get(seg)
+                    else:
+                        annotation = annotation.__annotations__.get(seg)
+                if have_annotation_for_attribute:
+                    check_type_internal(val, annotation, memo)
             except TypeCheckError as exc:
                 qualname = qualified_name(val, add_class_prefix=True)
                 exc.append_path_element(f"value assigned to {varname} ({qualname})")
--- a/tests/test_warn_on_error.py
+++ b/tests/test_warn_on_error.py
@@ -1,5 +1,5 @@
-from typing import List
 import os
+from typing import List
 
 import pytest
 
--- a/tests/test_warn_on_error.py
+++ b/tests/test_warn_on_error.py
@@ -1,4 +1,5 @@
 from typing import List
+import os
 
 import pytest
 
@@ -10,7 +11,12 @@ def test_check_type(recwarn):
         check_type(1, str, typecheck_fail_callback=warn_on_error)
 
     assert len(warning.list) == 1
-    assert warning.list[0].filename == __file__
+    if (
+        os.name == "nt"
+    ):  # file path name (and drive letter) is case insensitive on windows
+        assert warning.list[0].filename.lower() == __file__.lower()
+    else:
+        assert warning.list[0].filename == __file__
     assert warning.list[0].lineno == test_check_type.__code__.co_firstlineno + 2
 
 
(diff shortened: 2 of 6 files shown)


fefe982:
>  1.    Attribute is only handled in tuple, but not bare attribute assignments. Add check for bare attribute assignment
>  1.    `exp.id` is used for `Attribute`, but for `Attribute`, it points to the parent object, not the attribute. Changed it to use name ，which is obj.attr, the full path.

The new commit reverted change 2, keeping the annotation of the parent object. Instead, it changed `check_variable_assignment`, to get the real annotation of the attribute from the parent object. In this way the type of attribute can be correctly checked.

Other changes are due to the pytest does not pass on my Windows machine occasionally, as the path is case insensitive on Windows, and the case changed.

agronholm:
I'll try to find time to review this on the weekend.

tests/test_warn_on_error.py
```diff
@@ -10,7 +11,12 @@ def test_check_type(recwarn):
         check_type(1, str, typecheck_fail_callback=warn_on_error)
 
     assert len(warning.list) == 1
-    assert warning.list[0].filename == __file__
+    if (
+        os.name == "nt"
+    ):  # file path name (and drive letter) is case insensitive on windows
+        assert warning.list[0].filename.lower() == __file__.lower()
+    else:
+        assert warning.list[0].filename == __file__
```
agronholm, line 19: What exactly prompted this change? Was something failing?

src/typeguard/_functions.py
```diff
@@ -280,7 +281,20 @@ def check_variable_assignment(
...
+                have_annotation_for_attribute = True
+                for seg in varpath[1:]:
+                    if not inspect.isclass(annotation):
+                        have_annotation_for_attribute = False
+                        break
+                    if sys.version_info >= (3, 10):
+                        annotation = inspect.get_annotations(annotation).get(seg)
+                    else:
+                        annotation = annotation.__annotations__.get(seg)
+                if have_annotation_for_attribute:
```
agronholm, line 296: Separate unrelated control blocks for code readability. I wish I could enforce this with Ruff.
```suggestion

                have_annotation_for_attribute = True
                for seg in varpath[1:]:
                    if not inspect.isclass(annotation):
                        have_annotation_for_attribute = False
                        break

                    if sys.version_info >= (3, 10):
                        annotation = inspect.get_annotations(annotation).get(seg)
                    else:
                        annotation = annotation.__annotations__.get(seg)

                if have_annotation_for_attribute:
```

agronholm (changes requested):
Style fixes and a question.

agronholm:
So the thing is, Typeguard so far hasn't checked attribute assignments, or at least that it didn't intend to. I haven't yet fully delved into the details of this PR, but is your intention to add checks for attribute assignments?

agronholm:
> So the thing is, Typeguard so far hasn't checked attribute assignments, or at least that it didn't intend to. I haven't yet fully delved into the details of this PR, but is your intention to add checks for attribute assignments?

Never mind, judging from the tests it looks like this was indeed your intention. But attribute assignment is much more complicated than local variable assignment due to the potential interference by descriptors, and I'm not prepared to handle all that complexity. Until there is a plan to handle that aspect of attribute assignment, I'm not ready to accept such a new feature. Let's just focus on fixing this one false positive.

agronholm:
Ok, I've pushed a fix for the linked issue. Given that this PR (as I understand) expands the capabilities of Typeguard, we should take our time to sort out any potential issues with descriptors and `__setattr__()`, okay?

tests/test_warn_on_error.py
fefe982: It break test on Windows. The drive letter (`D:\`) can be either lower or upper case. And they do appear differently two sides sometime.

However, they do not always appear differently. So this sometimes make the tests fail on my windows machine. I do know how to definitely reproduce this. Windows is case insensitive. Maybe `os.path.normcase` can be used avoid the test for OS.

fefe982:
I did wonder why the code some check for `Attribute` and included  it, but not actually do the check :)

I only added type check for attributes that there is an annotation. Anything other than that are not checked.

I agree with that we leave this until we know more.

agronholm:
Oh, right, the issue this fixes was already dealt with. Closing.

PR #511, opened 2025-02-13
fix #510

### LINKED ISSUE #510: Problem with := operator
### Things to check first

### Typeguard version

4.4.1

### Python version

3.12.1

### What happened?

```
  File "test.py", line 3, in <module>
    import A
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/typeguard/_importhook.py", line 98, in exec_module
    super().exec_module(module)
  File "A.py", line 7, in <module>
    f()
  File "A.py", line 3, in f
    if v := tuple():
       ^^^^^^^^^^^^
TypeError: check_variable_assignment() takes 3 positional arguments but 4 were given
```

`test.py`:
```python3
with install_import_hook(('A',)):
    import A
```
`A.py`:
```python3
def f() -> None:
    v: str
    if v := tuple():
        pass
    print("OK")

f()
```
```shell
$ python3 A.py
OK

$ python3 test.py
<crash>
```
The problem only occurs when `v` is a local variable.

JohannesK71083:
## Changes

Fixes #510

In 889ad53 the parameters of check_variable_assignment in _functions.py were changed but one call in _transformer.py wasn't adjusted. Fixed by adjusting mentioned call and changing tests accordingly.
Furthermore, I had to change how the test handles mypy output because otherwise I couldn't get the tests to pass (even on the unmodified version). (see 5336b1c)

## Checklist

features)

### DIFF
--- a/tests/mypy/test_type_annotations.py
+++ b/tests/mypy/test_type_annotations.py
@@ -31,7 +31,7 @@ def get_negative_mypy_output() -> str:
     )
     output = process.stdout.decode()
     assert output
-    return re.sub(r'\n(?!\s|negative\.py)', ' ', output.replace("\r", ""))
+    return re.sub(r"\n(?!\s|negative\.py)", " ", output.replace("\r", ""))
 
 
 def get_expected_errors() -> Dict[int, str]:
--- a/tests/mypy/test_type_annotations.py
+++ b/tests/mypy/test_type_annotations.py
@@ -31,7 +31,7 @@ def get_negative_mypy_output() -> str:
     )
     output = process.stdout.decode()
     assert output
-    return output
+    return re.sub(r'\n(?!\s|negative\.py)', ' ', output.replace("\r", ""))
 
 
 def get_expected_errors() -> Dict[int, str]:
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -11,6 +11,7 @@ This library adheres to
   regression introduced in v4.4.1)
 - Fixed display of module name for forward references
   (`#492 <https://github.com/agronholm/typeguard/pull/492>`_; PR by @JelleZijlstra)
+- Fixed ``TypeError`` when using := (`#510 <https://github.com/agronholm/typeguard/issues/510>`_; PR: `#511 <https://github.com/agronholm/typeguard/pull/511>`_; PR by `@JohannesK71083 <https://github.com/JohannesK71083>`_)
 
 **4.4.1** (2024-11-03)
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -11,7 +11,8 @@ This library adheres to
   regression introduced in v4.4.1)
 - Fixed display of module name for forward references
   (`#492 <https://github.com/agronholm/typeguard/pull/492>`_; PR by @JelleZijlstra)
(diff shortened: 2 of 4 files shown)


agronholm:
I'll take a look at this over the weekend.

tests/mypy/test_type_annotations.py
```diff
@@ -31,7 +31,7 @@ def get_negative_mypy_output() -> str:
     )
     output = process.stdout.decode()
     assert output
-    return output
+    return re.sub(r"\n(?!\s|negative\.py)", " ", output.replace("\r", ""))
```
agronholm, line 34: I tested with the latest mypy locally and the tests passed just fine without this change. What error were you seeing?

agronholm: Oh, this seems to be a Windows-only problem. I was able to repro the failure on my Windows machine.

agronholm: I see...it inserts line breaks in unwanted positions and that's why it fails to match the output.

JohannesK71083: I don't think that it was a Windows only problem, because the Github action test also failed without it. I think it is due to a recent mypy update (that you maybe didn't pull yet?).

agronholm: I'm positive that I'm running the latest mypy (1.15.0) and I'm not seeing the problem on my Linux machine.

agronholm: I can, however, confirm that only mypy 1.15.0 fails on my Windows machine; earlier releases are fine.

JohannesK71083: I just reran the old version and it fails for that problem: https://github.com/JohannesK71083/typeguard/actions/runs/13313492169/job/37296369489#step:5:1

agronholm: Old version of typeguard you mean? This run installed mypy 1.15.0.

agronholm:
Thanks!

tests/mypy/test_type_annotations.py
JohannesK71083: > Old version of typeguard you mean? This run installed mypy 1.15.0.

Yes, I reran it with the old version of typeguard (and recent mypy version) on the linux GitHub action runner and it failed. Therefore, I don't think that it is a Windows problem. Not sure why it works at your local machine (local config, cache, bug???).

agronholm: I think mypy sniffs the terminal settings from an environment variable, and it can't do that on CI or Windows, so it defaults to a conversative 80 columns or something. But switching to JSON output should eliminate the problem entirely.

agronholm: OTOH, on Windows the terminal output, when I directly run mypy, seems just fine, not limited to 80 columns.

JohannesK71083: I think it has something to do with how the tests call mypy or something like that. (And that somehow changed because the Github tests used to pass). But I don't really have a clue why.

agronholm: OK, master is now using JSON output mode.

PR #534, opened 2025-10-05
small change: add PyPI link in README via badge

joshdavham:
I was checking out this repo earlier but couldn't find a PyPI link. I thought it would be a good idea to include one.

Lmk if you have any questions 👍

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,6 @@
+.. image:: https://img.shields.io/pypi/v/typeguard
+  :target: https://pypi.org/project/typeguard/
+  :alt: PyPI Link
 .. image:: https://github.com/agronholm/typeguard/actions/workflows/test.yml/badge.svg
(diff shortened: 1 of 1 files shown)


agronholm:
Why is a PyPI link necessary? You can easily find the project on PyPI, and there's really nothing there you would want that isn't already on GitHub.

joshdavham:
> Why is a PyPI link necessary?

It's not necessary.

> You can easily find the project on PyPI, and there's really nothing there you would want that isn't already on GitHub.

I'm generally quite paranoid about namesquatting and so before, I install a new python package, I like to see that there's a link to that PyPI package from the source repo to confirm they're the same project. This repo did not have such a link.

agronholm:
Or, maybe you can start from the PyPI and then check the verified details on the left?
image

agronholm:
At any rate, I don't see value in this so I'm closing the PR.

PR #551, opened 2026-05-01
Fixed crash in protocol check when subject method has no positional params

### LINKED ISSUE #550: IndexError: pop from empty list in check_signature_compatible when subject method has zero positional params
### Things to check first

### Typeguard version

4.5.1

### Python version

3.14.4

### What happened?

`typeguard.check_type(value, SomeProtocol)` crashed with `IndexError: pop from empty list` instead of returning or raising `TypeCheckError`.

```
File ".../typeguard/_checkers.py", line 769, in check_signature_compatible
    subject_args.pop(0)
IndexError: pop from empty list
```

The two `pop(0)` calls in `check_signature_compatible` :

https://github.com/agronholm/typeguard/blob/ff2b88b1c65f8828a178b4b2fffacd529cc76b26/src/typeguard/_checkers.py#L764-L769

If `inspect.signature(subject_attr)` reports zero positional-or-keyword params and no `*args`, `subject_args` is empty and the pop blows up. Callable instances, partials, and other non-staticmethod/non-classmethod descriptors all hit this path.

What I expected was a `TypeCheckError`. The same arity mismatch with one more positional arg on the subject raises a clean `TypeCheckError` ("too few positional arguments"). Drop the count by one and it becomes `IndexError`. Equivalent mismatch, two different exception types, only one of which `check_type` is documented to raise, so callers catching `TypeCheckError` leak the `IndexError` through.

The fix I am running locally is to guard each pop:

```python
if protocol_type == "instance" and protocol_args:
    protocol_args.pop(0)
if subject_type == "instance" and subject_args:
    subject_args.pop(0)
```

With that, the repro raises `TypeCheckError` and the rest of the test suite still passes for me.

Happy to PR with a regression test under `tests/test_checkers.py::TestProtocol`. If you would rather the empty case raise an explicit `TypeCheckError` earlier with a more specific message, let me know and I will rework.

Stdlib only, no third-party deps:

```python
from typing import Protocol

class _ZeroArg:
    def __call__(self):
        return None

class Subject:
    method = _ZeroArg()

class P(Protocol):
    def method(self, x: int) -> None: ...

check_type(Subject(), P)
# IndexError: pop from empty list
```

Run on Python 3.14, typeguard 4.5.1.

HackedRico:
## Changes

Fixes #550

`check_signature_compatible` in `src/typeguard/_checkers.py` calls `.pop(0)` on `protocol_args` and `subject_args` to strip the implicit `self`, but does so unconditionally whenever the corresponding side is classified as `"instance"`. If the subject method's `inspect.signature(...)` reports zero positional-or-keyword parameters and no `*args` (e.g. a callable instance assigned at class scope, a `functools.partial` with all positional args bound, or a method proxied by a code-generation framework), `subject_args` is empty and `pop(0)` raises `IndexError` instead of letting the function fall through to its existing `zip_longest` comparison and emit a `TypeCheckError`.

This guards each pop with a non-empty check. After the change, the previously-crashing input raises `TypeCheckError("... has too few positional arguments")` along the same path as the equivalent N → N-1 mismatch with one more arg, restoring `check_type`'s documented contract that callers can catch via `except TypeCheckError`.

Regression test added under `tests/test_checkers.py::TestProtocol::test_subject_method_no_positional_params`.

## Checklist

features)

these instructions.

### Updating the changelog

- Fixed crash in protocol check when subject method has no positional params (`#550 <https://github.com/agronholm/typeguard/issues/550>`_; PR by @HackedRico)

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -4,6 +4,12 @@ Version history
 This library adheres to
 `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**UNRELEASED**
+
+- Fixed ``IndexError`` raised from ``check_signature_compatible`` when the subject
+  method has no positional parameters
+  (`#550 <https://github.com/agronholm/typeguard/issues/550>`_; PR by @HackedRico)
+
 **4.5.1** (2026-02-19)
 
 - Fixed iterable unpacking incorrectly calculating the cut-off offset of the item list
--- a/src/typeguard/_checkers.py
+++ b/src/typeguard/_checkers.py
@@ -761,11 +761,11 @@ def check_signature_compatible(subject: type, protocol: type, attrname: str) -> 
         ]
 
         # Remove the "self" parameter from the protocol arguments to match
-        if protocol_type == "instance":
+        if protocol_type == "instance" and protocol_args:
             protocol_args.pop(0)
 
         # Remove the "self" parameter from the subject arguments to match
-        if subject_type == "instance":
+        if subject_type == "instance" and subject_args:
             subject_args.pop(0)
 
         for protocol_arg, subject_arg in zip_longest(protocol_args, subject_args):
--- a/tests/test_checkers.py
+++ b/tests/test_checkers.py
@@ -1405,8 +1405,8 @@ class TestProtocol:
         )
 
     def test_subject_method_no_positional_params(self) -> None:
-        class _ZeroArg:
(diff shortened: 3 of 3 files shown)


agronholm:
You checked the checkbox for adding a new changelog entry when you clearly haven't done so. What gives?

HackedRico:
Fixed

agronholm:
Thanks!

PR #554, opened 2026-05-20
Fix compatibility with Python 3.15

hrnciar:
- Drop removed debug_override parameter from cache_from_source() call
- Accept extra positional args in source_to_code() for new fullname parameter
- Guard no_type_check_decorator import behind version check (removed in 3.15)

Assisted-by: Cursor

## Changes

Fixes #. 

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -4,6 +4,11 @@ Version history
 This library adheres to
 `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**UNRELEASED**
+
+- Compatibility with Python 3.15, drop removed debug_override parameter from cache_from_source()
+  call (`#554 <https://github.com/agronholm/typeguard/pull/554>`_; PR by @hrnciar)
+
 **4.5.2** (2026-05-14)
 
 - Fixed ``IndexError`` raised from ``check_signature_compatible`` when the subject
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,8 +6,9 @@ This library adheres to
 
 **UNRELEASED**
 
-- Compatibility with Python 3.15, drop removed debug_override parameter from cache_from_source()
-  call (`#554 <https://github.com/agronholm/typeguard/pull/554>`_; PR by @hrnciar)
+- Fixed compatibility with Python 3.15, drop removed debug_override parameter from
+  cache_from_source() call
+  (`#554 <https://github.com/agronholm/typeguard/pull/554>`_; PR by @hrnciar)
 
 **4.5.2** (2026-05-14)
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,8 +6,7 @@ This library adheres to
 
 **UNRELEASED**
 
-- Fixed compatibility with Python 3.15, drop removed debug_override parameter from
-  cache_from_source() call
+- Fixed compatibility with Python 3.15
   (`#554 <https://github.com/agronholm/typeguard/pull/554>`_; PR by @hrnciar)
 
 **4.5.2** (2026-05-14)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: 3.14",
+    "Programming Language :: Python :: 3.15",
 ]
 requires-python = ">= 3.9"
 dependencies = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,7 @@ pretty = true
 files = ["src"]
 
 [tool.tox]
-env_list = ["py39", "py310", "py311", "py312", "py313", "py314"]
+env_list = ["py39", "py310", "py311", "py312", "py313", "py314", "py315"]
 skip_missing_interpreters = true
 requires = ["tox >= 4.22"]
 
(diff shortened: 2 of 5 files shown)


agronholm:
To prove 3.15 compatibility, this should add 3.15 to the test matrix.

src/typeguard/_importhook.py
```diff
@@ -48,16 +48,16 @@ def _call_with_frames_removed(
     return f(*args, **kwargs)
 
 
-def optimized_cache_from_source(path: str, debug_override: bool | None = None) -> str:
-    return cache_from_source(path, debug_override, optimization=OPTIMIZATION)
+def optimized_cache_from_source(path: str, **kwargs: object) -> str:
+    return cache_from_source(path, optimization=OPTIMIZATION)
```
agronholm: You're just dropping `kwargs` here?

docs/versionhistory.rst
```diff
@@ -4,6 +4,10 @@ Version history
 This library adheres to
 `Semantic Versioning 2.0 <https://semver.org/#semantic-versioning-200>`_.
 
+**UNRELEASED**
+
+- Compatibility with Python 3.15, drop removed debug_override parameter from cache_from_source() call; PR by @hrnciar
```
agronholm: Please respect the instructions in the PR template.

tests/dummymodule.py
```diff
@@ -42,6 +41,7 @@
 
 
 if sys.version_info <= (3, 13):
+    from typing import no_type_check_decorator
```
agronholm, line 44: Why did you move this import?

src/typeguard/_importhook.py
```diff
@@ -48,16 +48,16 @@ def _call_with_frames_removed(
...
 
 
 class TypeguardLoader(SourceFileLoader):
     @staticmethod
     def source_to_code(
         data: Buffer | str | ast.Module | ast.Expression | ast.Interactive,
         path: Buffer | str | PathLike[str] = "<string>",
-        *,
+        *args: object,
```
agronholm: Why should this accept arbitrary positional arguments that are then dropped on the floor?

agronholm (changes requested):
This isn't looking good, but I'm willing to give you a chance to fix it.

agronholm:
Please do not force push your changes. If you do, I won't know what you changed in response to my feedback.

hrnciar:
I am sorry, I was about to reply in the above conversation.

src/typeguard/_importhook.py
hrnciar: This was a mistake, I should just drop the debug_override. The argument was removed in 3.15 and since it is always None it shouldn't be missing on the older pythons.

tests/dummymodule.py
hrnciar: `no_type_check_decorator` was removed in 3.15, technically it could be guarded by `sys.version_info <= (3, 14)`, but since it is already only used when python <= 3.13 I just moved the import there.

agronholm:
You added 3.15 to `tox` configuration but it's still missing from the Github Actions workflow.

src/typeguard/_importhook.py
hrnciar: Right, I should have added just the new positional `fullname` argument.

tests/dummymodule.py
agronholm: Makes sense.

agronholm:
Alright, getting better! Is there any place where the new `fullname` argument could be passed to?

hrnciar:
From the 3.15 documentation:
> Many functions related to compiling or parsing Python code, such as [compile()](https://docs.python.org/3.15/library/functions.html#compile), [ast.parse()](https://docs.python.org/3.15/library/ast.html#ast.parse), [symtable.symtable()](https://docs.python.org/3.15/library/symtable.html#symtable.symtable), and [importlib.abc.InspectLoader.source_to_code()](https://docs.python.org/3.15/library/importlib.html#importlib.abc.InspectLoader.source_to_code), now allow the module name to be passed. It is needed to unambiguously [filter](https://docs.python.org/3.15/library/warnings.html#warning-filter) syntax warnings by module name. (Contributed by Serhiy Storchaka in [gh-135801](https://github.com/python/cpython/issues/135801).)

In our case, it could be used in the two `_call_with_frames_removed` calls (ast.parse and compile).

agronholm:
> From the 3.15 documentation:
> 
> > Many functions related to compiling or parsing Python code, such as [compile()](https://docs.python.org/3.15/library/functions.html#compile), [ast.parse()](https://docs.python.org/3.15/library/ast.html#ast.parse), [symtable.symtable()](https://docs.python.org/3.15/library/symtable.html#symtable.symtable), and [importlib.abc.InspectLoader.source_to_code()](https://docs.python.org/3.15/library/importlib.html#importlib.abc.InspectLoader.source_to_code), now allow the module name to be passed. It is needed to unambiguously [filter](https://docs.python.org/3.15/library/warnings.html#warning-filter) syntax warnings by module name. (Contributed by Serhiy Storchaka in [gh-135801](https://github.com/python/cpython/issues/135801).)
> 
> In our case, it could be used in the two `_call_with_frames_removed` calls (ast.parse and compile).

If you could do that on 3.15, that'd be great.

pyproject.toml
agronholm, line 24:
```suggestion
    "Programming Language :: Python :: 3.14",
    "Programming Language :: Python :: 3.15",
```

agronholm:
Looks like I introduced mypy errors with my simplification. Those need to be undone. Otherwise this looks good now!

hrnciar:
Thank you for the help!

agronholm (approved):
Thanks!

PR #556, opened 2026-06-18
fix: set __wrapped__ on instrumented functions to preserve signature

### LINKED ISSUE #553: Title: @typechecked decorator does not preserve function signature, causing false "unexpected argument" errors in PyCharm
### Things to check first

### Typeguard version

4.x

### Python version

3.13

### What happened?

**Description:**
Since recent versions of PyCharm have improved their static type analysis, the @typechecked decorator from typeguard v3/v4 causes false "unexpected argument" errors for all parameters of decorated functions.

**Root Cause:**
In typeguard v2, @typechecked used functools.wraps internally, which preserved the original function's __wrapped__ metadata and signature. PyCharm (and other static analysis tools) could therefore look through the wrapper and inspect the original signature correctly.
In typeguard v3+, the decorator generates a wrapper that does not fully preserve the original signature in a way that PyCharm's type checker can resolve. As a result, PyCharm treats the wrapped function as having no defined parameters, and reports every argument passed to it as "unexpected".

**Expected behavior:**
@typechecked should preserve the original function signature so that static analysis tools can inspect it correctly — as was the case in typeguard v2.

**Suggested fix:**
Apply functools.update_wrapper (or functools.wraps) inside the decorator implementation to ensure __wrapped__, __annotations__, and __signature__ are correctly propagated to the wrapper function.

use the current pycharm version with a project where typeguard is used.

gaoflow:
The `@typechecked` decorator replaces the original function with an
AST-instrumented copy but did not set `__wrapped__` on the new
function object. This prevented static analysis tools (e.g. PyCharm,
mypy) from inspecting the original function signature, causing false
"unexpected argument" errors.

Fix: use `functools.update_wrapper(new_function, f, updated=())` in
`instrument()` to set `__wrapped__` and copy `__dict__` without
overwriting the already-manually-copied attributes.

Fixes: #553


agronholm:
More AI slop. Banned.

PR #558, opened 2026-06-30
Fix annotated walrus assignment handling (#557)

### LINKED ISSUE #557: Mishandled of annotated walrus operator
### Things to check first

### Typeguard version

4.5.2

### Python version

3.10 and 3.12

### What happened?

"""

                                             corrupting the variable.

"""

@typechecked
def case_none() -> None:
    x: int | None = 1
    if (x := None) is not None:
        pass

@typechecked
def case_int() -> None:
    x: int = 0
    if (x := 5) > 0:
        pass

@typechecked
    if (x := None) is None:
        pass

@typechecked
    x: str = "init"
    if (x := "abc"):
        pass

@typechecked
    biggest: int = 0
    current: int = 0
    return biggest

def _run(label, fn, *args):
    try:
        result = fn(*args)
    except Exception as exc:
    else:

if __name__ == "__main__":

HarperZ9:
Fixes #557.

`visit_NamedExpr` wrapped the `check_variable_assignment` target in a doubly-nested list `[[('x', T)]]`, whereas `visit_AnnAssign` and `visit_AugAssign` emit a single-level `[('x', T)]`. At runtime `check_variable_assignment` treats any `groups` element that is a `list` as a tuple-unpacking target (calling `list(value)`). With the extra nesting, an annotated variable assigned via the walrus operator was mishandled:

- a non-iterable value raised a spurious `TypeError: '<type>' object is not iterable`
- an iterable value (e.g. `str`) was silently coerced to a `list`

Repro on `master`:

```python

@typechecked
def f():
    x: int
    if (x := 5) > 100:
        pass
    return x

f()  # TypeError: 'int' object is not iterable
```

### Change
- Remove the extra `List` nesting in `visit_NamedExpr` so it matches the other assignment visitors.
- Update the two `test_transformer` assertions that encoded the buggy `[[('x', int)]]` output.
- Add runtime regression tests (`TestAssignmentExpr`): the non-iterable case no longer raises, the iterable case is not coerced, and a type mismatch on the walrus target still raises `TypeCheckError`.
- Add a changelog entry.

`tests/test_transformer.py` and `tests/test_instrumentation.py` pass locally; `ruff check` and `ruff format --check` are clean on the changed files.


HarperZ9:
A note on the single red check: the Python 3.11 test suite itself passed in CI:

`503 passed, 23 skipped, 9 xfailed in 10.61s`

The job failed afterward while posting coverage to Coveralls. The relevant tail is a Coveralls/Cloudflare response:

`Error: Bad Gateway (502)` from `https://coveralls.io/api/v1/jobs`.

I tried to rerun the failed job, but GitHub requires repository admin rights for that action, so I cannot clear it from the fork side. No code change appears to be needed for this failure.

agronholm:
Handled via 32e7c85f74d12be72acd5856869e5110a7d4549e.

PR #562, opened 2026-07-08
Fix false positive checking a class against type[Protocol] with instance attributes

### LINKED ISSUE #499: Class is not compatible with protocol because it has no attribute named ...
### Things to check first

### Typeguard version

4.4.1

### Python version

3.13.0

### What happened?

I think https://github.com/agronholm/typeguard/issues/498 is not fully fixed yet.

`bug.py`
```python

install_import_hook()

from code import Class, function_of_instance, function_of_type

# This works:
function_of_instance(Class())

# This does not:
function_of_type(Class)
```

`code.py`
```python
from typing import Protocol

class Proto(Protocol):
    foo: str

class Class(Proto):
    def __init__(self):
        self.foo = "bar"

def function_of_instance(instance: Proto): ...

def function_of_type(type_: type[Proto]): ...
```

AnayGarodia:
Fixes #499

Checking a class (not an instance) against `type[SomeProtocol]` raised a false-positive `TypeCheckError` when the protocol declared a non-`ClassVar` attribute that the class sets in `__init__`:

```python
class Proto(Protocol):
    foo: str

class Foo:
    def __init__(self) -> None:
        self.foo = "bar"

check_type(Foo, type[Proto])   # raised, though mypy/pyright accept it
```

Per PEP 544, an attribute annotation without `ClassVar` describes an *instance* attribute, so a class object is not expected to carry it. This implements the approach you outlined in the issue: when the subject is a class, `check_protocol` now ignores non-`ClassVar` attribute annotations and only requires the `ClassVar` ones to be present on the class. Instance checks are unchanged, and method/callable checks are untouched.

Added two regression tests: a class satisfying a protocol via an instance attribute no longer raises, and a class missing a required `ClassVar` member still does. The full `tests/test_checkers.py` suite passes.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,10 +6,6 @@ This library adheres to
 
 **UNRELEASED**
 
-- Fixed false positive when checking a class against ``type[SomeProtocol]`` where the
-  protocol declares non-``ClassVar`` (instance) attributes; only ``ClassVar`` members
-  are now required on the class itself
-  (`#499 <https://github.com/agronholm/typeguard/issues/499>`_)
 - Fixed compatibility with Python 3.15
   (`#554 <https://github.com/agronholm/typeguard/pull/554>`_; PR by @hrnciar)
 - Fixed an assignment expression against an annotated name (``x: int``, then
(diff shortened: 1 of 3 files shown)


agronholm:
Thanks!

PR #563, opened 2026-07-18
Fix annotated assignment expressions being treated as unpacking targets

### LINKED ISSUE #557: Mishandled of annotated walrus operator

dchaudhari7177:
## Summary

Fixes #557. `visit_NamedExpr` wrapped its single target in an extra list, emitting

```python
check_variable_assignment(otherfunc(), [[('x', int)]], memo)
```

where the plain-assignment path emits `[('x', int)]` for a single name. `check_variable_assignment` reads a **list** group as a tuple-unpacking target and a **tuple** group as a single target, so an annotated walrus took the unpacking branch and consumed the assigned value with `list(value)`.

Two wrong outcomes, both only when the name carries an annotation:

| assigned value | before | after |
| --- | --- | --- |
| `5` | `TypeError: 'int' object is not iterable` | `5` |
| `None` | `TypeError: 'NoneType' object is not iterable` | `None` |
| `"abc"` | **no error**, `x` becomes `['a', 'b', 'c']` | `"abc"` |
| `"not an int"` (wrong type) | `TypeCheckError` | `TypeCheckError` (unchanged) |

The third row is the dangerous one — no exception, just a silently corrupted variable.

A walrus target is always a single `Name`, so it now emits a bare tuple like the assignment path does. Type checking itself is unaffected: a genuinely wrong value still raises `TypeCheckError`.

## Tests

New `TestAssignmentExpression` in `test_typechecked.py` covers the behaviour rather than the emitted source: non-iterable value, iterable value not consumed, an annotated *argument* reused in a walrus, an iterator not drained, and a wrong-typed value still raising.

Reverting only `src/` fails 4 of the 5 (the 5th passes either way, since it asserts the error that already worked):

```
FAILED TestAssignmentExpression::test_non_iterable_value
FAILED TestAssignmentExpression::test_value_is_not_consumed
FAILED TestAssignmentExpression::test_annotated_argument
FAILED TestAssignmentExpression::test_iterator_value_is_not_consumed
```

`test_transformer.py::TestAssign::test_assignment_expr` and `test_assignment_expr_annotated_argument` asserted the old `[[('x', int)]]` output, i.e. they encoded the defect, so they now expect the single-target form. Worth noting those two golden tests passed throughout the bug's lifetime — pinning emitted source didn't catch it, which is why the new tests assert runtime behaviour instead.

## Verification

- Full suite (`--ignore=tests/mypy`) → **516 passed, 7 skipped, 9 xfailed**
- `ruff check src tests` → all checks passed; `ruff format --check` → 31 files already formatted
- Added an UNRELEASED entry to `docs/versionhistory.rst`

`tests/mypy` is excluded above: both tests there fail with `FileNotFoundError` on a clean tree too (no mypy binary in my environment), unrelated to this change.


src/typeguard/_transformer.py
```diff
@@ -1148,15 +1148,14 @@ def visit_NamedExpr(self, node: NamedExpr) -> Any:
                 func_name,
                 [
                     node.value,
+                    # A walrus target is a single Name, so the group is a bare
+                    # tuple. Wrapping it in a list marks it as an unpacking
+                    # target, which makes check_variable_assignment consume the
+                    # value with list().
```
agronholm: It's enough to explain this in the PR. Leaving noisy comments behind is unwelcome.
```suggestion
```

dchaudhari7177:
Removed, thanks — the reasoning lives in the PR description instead.

For reference, the emitted call goes from `[[('x', int)]]` to `[('x', int)]`: the inner list marked the group as an unpacking target, so `check_variable_assignment` consumed the value with `list()`.

`pytest tests/test_transformer.py` still passes (84 passed, 1 skipped). The 2 failures in the full run are `tests/mypy/` looking for a mypy binary that isn't installed here.

PR #564, opened 2026-07-18
Fix walrus (:=) assignments to annotated variables mishandled as tuple-unpacking

### LINKED ISSUE #557: Mishandled of annotated walrus operator

agu2347:
Fixes #557.

`TypeguardTransformer.visit_NamedExpr()` built the `check_variable_assignment()` target argument as a doubly-nested list: `[[('x', annotation)]]`. Every other single-target call site (e.g. `visit_AnnAssign()`, used for plain `x: T = value`) builds it as a single-nested list: `[('x', annotation)]`.

`check_variable_assignment()` distinguishes single-target from tuple-unpacking assignments purely by whether each element of the outer `groups` list is itself a list (unpacking) or a bare tuple (single target). The extra list layer around the walrus operator's target therefore made every walrus assignment to an annotated variable look like a tuple-unpacking assignment, triggering `list(value)` on the right-hand side:

- Non-iterable values (`None`, `int`, ...) raised a spurious internal `TypeError` instead of being type-checked normally.
- Iterable values (`str`, `list`, ...) didn't raise, but the assigned variable was silently replaced by `list(value)` -- e.g. a string `"abc"` became `['a', 'b', 'c']` -- corrupting the variable's value without any error being surfaced.

As diagnosed precisely in the issue.

**Fix**: remove the extra `List` wrapper so the walrus operator's single target is built the same way as every other single-target call site.

**Testing**: added runtime regression tests covering the exact scenarios from the issue -- a non-iterable value (previously a spurious `TypeError`), an iterable value (previously silent corruption), and confirmation that an actually-invalid assignment (`None` to an `int`-annotated variable) now correctly raises `TypeCheckError` instead of an unrelated internal `TypeError`. I confirmed both new tests fail with the original code and pass with the fix.

I also updated two existing AST-snapshot tests in `test_transformer.py` that were asserting the old, incorrect `[[('x', int)]]` structure as expected output -- they were unknowingly encoding this bug into the test suite. The analogous plain-assignment test in the same file already asserts the correct `[('x', int)]` format, confirming this was purely a walrus-specific inconsistency.

Ran the full existing test suite: 396 passed (394 baseline + 2 new), 3 skipped, 3 xfailed, matching the baseline exactly aside from the new tests. The 4 remaining failures (2 missing `mypy` binary, 2 pytest deprecation-warning related) are present identically on a clean checkout of `master` and are unrelated environment/dependency issues in my sandbox.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,6 +6,11 @@ This library adheres to
 
 **UNRELEASED**
 
+- Fixed walrus (``:=``) assignments to an annotated variable being instrumented
+  as if they were tuple-unpacking assignments, causing a spurious ``TypeError``
+  for non-iterable values and silent corruption (via ``list(value)``) of
+  iterable values
+  (`#557 <https://github.com/agronholm/typeguard/issues/557>`_)
 - Fixed compatibility with Python 3.15
   (`#554 <https://github.com/agronholm/typeguard/pull/554>`_; PR by @hrnciar)
 - Dropped support for Python 3.9
--- a/src/typeguard/_transformer.py
+++ b/src/typeguard/_transformer.py
@@ -1150,13 +1150,8 @@ class TypeguardTransformer(NodeTransformer):
                     node.value,
                     List(
                         [
-                            List(
-                                [
-                                    Tuple(
-                                        [Constant(node.target.id), annotation],
-                                        ctx=Load(),
-                                    )
-                                ],
+                            Tuple(
+                                [Constant(node.target.id), annotation],
                                 ctx=Load(),
                             )
                         ],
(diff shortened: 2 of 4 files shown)


agronholm:
Is this different from #563 then?

agronholm:
Closing as duplicate. Why didn't you check for existing PRs? This just forces me to spend additional time doing nothing useful.

PR #565, opened 2026-07-19
Do not erase a meaningful empty tuple[()] subscript in the transformer

chuenchen309:
## Summary

The empty-tuple annotation `tuple[()]` is silently erased to bare `tuple` by the AST transformer, so the **import hook** and **`@typechecked`** (which share the transformer) wrongly accept *any* tuple for a `tuple[()]` parameter. `check_type` — which does not go through the transformer — correctly rejects it. Two of typeguard's three checking paths disagree with the third, and the third is the correct one.

```python
from typeguard import typechecked, check_type

check_type((1,), tuple[()])          # correctly raises "tuple is not an empty tuple"

@typechecked
def f(x: tuple[()]): ...
f((1,))                              # BUG: accepted (should raise); same via install_import_hook
```

| path | before | after |
|---|---|---|
| `check_type` | reject (correct) | reject |
| `@typechecked` (source annotation) | **accept — wrong** | reject |
| import hook | **accept — wrong** | reject |

## Root cause

`src/typeguard/_transformer.py`, `AnnotationTransformer.visit_Subscript`:

```python
# If all items in the subscript were Any, erase the subscript entirely
if all(item is None for item in items):
    return node.value
```

For `tuple[()]` the AST slice is `Tuple(elts=[])`, so `items == []` and `all([])` is vacuously `True` — the transformer rewrites `tuple[()]` to `tuple`, dropping the constraint. (The instrumented source confirms it: the annotation is emitted as `{'x': (x, tuple)}` with the `[()]` gone.) typeguard's checker itself has explicit empty-tuple support (`_checkers.py`, "is not an empty tuple"), so the intended verdict is unambiguous.

## Fix

Guard the erasure so it only fires for a *non-empty* all-`Any` subscript:

```python
# If all items in the subscript were Any, erase the subscript entirely
# (but not for an empty subscript like tuple[()], which is meaningful)
if items and all(item is None for item in items):
    return node.value
```

Genuine all-`Any` erasure (e.g. `dict[Any, Any]`) is unchanged.

## Verification

- New `test_empty_tuple` / `test_empty_tuple_fail` in `tests/test_instrumentation.py` run under **both** the `typechecked` and `importhook` parametrizations. On the unfixed tree `test_empty_tuple_fail` fails for both (a non-empty tuple is wrongly accepted); with the fix all 4 pass.
- `test_transformer.py test_instrumentation.py test_checkers.py test_typechecked.py`: **495 passed, 7 skipped, 9 xfailed**, no failures. Full `tests/`: 523 passed (the only 2 failures are `tests/mypy/` `FileNotFoundError: 'mypy'` — mypy not installed in my env, unrelated).
- Changelog entry added in `docs/versionhistory.rst`.

---

*This PR was authored by an AI coding agent (Claude Code) running on this account: the AI found the bug, ran the repro, wrote the tests, and wrote this description. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.*


docs/versionhistory.rst
```diff
@@ -20,6 +20,9 @@ This library adheres to
   protocol declares non-``ClassVar`` (instance) attributes; only ``ClassVar`` members
   are now required on the class itself
   (`#499 <https://github.com/agronholm/typeguard/issues/499>`_)
+- Fixed the import hook/instrumenter erasing the subscript of an empty tuple annotation
+  (``tuple[()]``) down to a bare ``tuple``, causing any tuple (including non-empty ones)
+  to pass the check
```
agronholm, line 29: The changelog entry is missing the required pull request link.

agronholm (approved):
LGTM!

Issue #233, opened 2022-02-23
0 and 1 are accepted as `True` and `False` to `Literal[...]`


mhostetter:
I discovered that if `True` is included in a `Literal` list, the integer 1 will be accepted as "True". The same goes for 0 and "False".

## Minimal Example

```python
# bug.py
import typeguard

from typing_extensions import Literal

@typeguard.typechecked
def function(arg: Literal[True, False, "red", "blue"]):
    print(arg)
```

## Output

```python
In [1]: %run bug.py

In [2]: function("red")
red

In [3]: function(True)
True

In [4]: function(1)
1

In [5]: function(0)
0
```

## Expected Output

```python
In [6]: function(1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-4c0fe5c458a0> in <module>
----> 1 function(1)

~/.local/lib/python3.8/site-packages/typeguard/__init__.py in wrapper(*args, **kwargs)
   1030     def wrapper(*args, **kwargs):
   1031         memo = _CallMemo(python_func, _localns, args=args, kwargs=kwargs)
-> 1032         check_argument_types(memo)
   1033         retval = func(*args, **kwargs)
   1034         try:

~/.local/lib/python3.8/site-packages/typeguard/__init__.py in check_argument_types(memo)
    873                 check_type(description, value, expected_type, memo)
    874             except TypeError as exc:  # suppress unnecessarily long tracebacks
--> 875                 raise TypeError(*exc.args) from None
    876
    877     return True

TypeError: the value of argument "arg" must be one of (True, False, 'red', 'blue'); got 1 instead
```

agronholm:
The literal checker checks by equality, and `1 == True` and `0 == False` in Python.

PR #566, opened 2026-07-23
Fix Literal check depending on argument order for bools

uttam12331:
## Changes

No existing issue — found while reading `check_literal`.

`check_literal` used `final_args.index(value)`, which returns the first `==` equal element. Because `bool` is a subclass of `int` (`1 == True`), `True` matched the `int` `1` in `Literal[1, True]`; the following `type(...) is type(...)` check then failed, and the genuinely matching `True` later in the tuple was never examined.

So the result depended on the order the arguments were written in:

```python
>>> check_type(True, Literal[1, True])
TypeCheckError: bool is not any of (1, True)   # rejected
>>> check_type(True, Literal[True, 1])         # same type, accepted
>>> check_type(False, Literal[0, False])
TypeCheckError: bool is not any of (0, False)  # rejected
>>> check_type(False, Literal[False, 0])       # same type, accepted
```

This scans the arguments instead, requiring both a type and a value match:

```python
for arg in final_args:
    if type(arg) is type(value) and arg == value:
        return
```

The type test is deliberately kept first, so it still short-circuits values whose `__eq__` does not return a bool — the case the previous `except ValueError` handler covered.

The int/bool separation added in #233 is preserved: `Literal[True]` still rejects `1`, `Literal[False]` still rejects `0`, and `Literal[1]` still rejects `True`.

## Checklist

features) — n/a, no documented behavior changes


agronholm (approved):
Thanks!

Issue #347, opened 2023-04-26
Emit a warning when instrumentation encounters a type it cannot check


agronholm:
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

The instrumentation code should emit a warning (when in debug mode) when it encounters a type that has been designated as "ignored" (due to, for example, being imported under `if TYPE_CHECKING:`.

### Use case

It may be sometimes difficult to understand why a type check has been omitted from the instrumented code. Adding code comments for this is not possible because comments are not part of the abstract syntax tree.

agronholm:
There is the option of storing the comments on the transformer itself, and injecting them into the debug output using less sophisticated methods like string searching. This is very difficult to do reliably though.

PR #570, opened 2026-08-18
Warn on type checks skipped due to TYPE_CHECKING-only imports

ChrisJr404:
This implements the enhancement requested in #347: surfacing type checks that get silently dropped because the annotation refers to a name that only exists during static type checking.

### Motivation

When an annotation references a name imported inside an `if TYPE_CHECKING:` block, that name doesn't exist at run time, so the instrumenter has to skip the check. Today that happens invisibly, and (as the issue notes) it can be hard to work out why a value you expected to be validated was left unchecked.

### What this does

When `debug_instrumentation` is enabled, the instrumenter now emits an `InstrumentationWarning` for each such skipped check, e.g.:

```
skipped type checking the annotation reference 'FooBar' in 'foo' because 'FooBar'
is only available during static type checking (e.g. imported in an
'if TYPE_CHECKING:' block)
```

It's gated on `debug_instrumentation` to match the issue ("when in debug mode") and to keep normal runs quiet, so it fits naturally alongside the existing debug output in both the import hook and `@typechecked`.

### Implementation notes

- The `AnnotationTransformer` already collapses an annotation to `None` at three leaf points (`visit_Name`, `visit_Attribute`, `visit_Subscript`) when it hits an ignored name. Since every annotation flows through `_convert_annotation`, those three spots are the single choke point where a skipped check can be observed, so the transformer records `(name, scope)` there.
- To keep the message accurate, recording is limited to names that were actually collected from a module-level `if TYPE_CHECKING:` block (tracked in a new `typechecking_only_names` set), so names ignored for unrelated reasons (parameter names, etc.) don't produce a misleading warning.
- The name-extraction logic that `is_ignored_name` used inline is factored out into a small `get_annotation_base_name` helper and reused, so there's no duplicated parsing.

### Tests / docs

- New transformer tests covering the collected `skipped_type_checks` set and the emitted warning.
- Updated `test_debug_instrumentation` to assert the new warning fires (the existing `dummymodule` fixture already has a `TYPE_CHECKING`-guarded hint).
- Added a user-guide note under "Debugging instrumented code" and a changelog entry.

Full test suite (including the mypy subprocess tests), `ruff check`, `ruff format --check`, and `mypy src/typeguard` all pass.


agronholm:
I'm favorable to adding a warning, but I feel like reusing that flag for this purpose is wrong. I would be open to adding a new configuration flag to address this need.

ChrisJr404:
Good call, reusing that flag wasn't right. I've moved the warning behind a dedicated `warn_on_unchecked_types` config option (defaults to `False`), so `debug_instrumentation` no longer controls it and the default behavior is unchanged. It's wired through the same way as the sibling options: the global config, the `@typechecked` override, and a matching `--typeguard-warn-on-unchecked-types` pytest flag. Docs, changelog and tests are updated, including one asserting it stays silent by default even with `debug_instrumentation` on.

I went with `warn_on_unchecked_types` as the name since you didn't specify one, but I'm happy to rename it if you'd prefer something else.

PR #571, opened 2026-08-21
Unwrap ReadOnly qualifier when checking TypedDict items

jaideeppyne:
## Problem

`check_typed_dict` unwraps the `NotRequired`/`Required` qualifiers before checking each item's value type, but not `ReadOnly` (PEP 705). A `ReadOnly[...]` item is left wrapped, so `get_origin()` doesn't match a concrete type and the value type is **never checked** — every value passes:

```python
from typing import TypedDict
from typing_extensions import ReadOnly

class TD(TypedDict):
    x: ReadOnly[int]

check_type({"x": "not an int"}, TD)   # passes — should raise TypeCheckError
```

`Required[ReadOnly[int]]` and `ReadOnly[NotRequired[int]]` are affected the same way. `ReadOnly` has been stable since Python 3.13 (and available on `typing_extensions` earlier).

## Fix

Replace the `if/elif` in the qualifier-stripping loop with a `while` loop that strips `NotRequired`/`Required`/`ReadOnly` in any nesting order, so the underlying type is checked. `ReadOnly` is imported the same way as the existing version-gated qualifiers (`typing` on 3.13+, `typing_extensions` below).

## Tests

Added `test_readonly_pass` / `test_readonly_fail` to `TestTypedDict`, parametrized over the `typing` and `typing_extensions` providers like the existing `NotRequired`/`Required` tests (skipping a provider that lacks `ReadOnly`). `test_readonly_fail` fails on `master` (the value silently passes) and passes with this change; the full `test_checkers.py` suite stays green (238 passed).

Also added a changelog entry.

---

*Disclosure: prepared with AI assistance and reviewed/verified by me before submission.*


agronholm (changes requested):
This PR implements an iterative unwrapping of relevant annotations, but the tests don't exercise that. Please add at least one test where nested `ReadOnly` and `Required` / `NotRequired` are used.

jaideeppyne:
Added `test_readonly_nested_qualifiers`: it covers `ReadOnly[Required[...]]` and `ReadOnly[NotRequired[...]]` plus the reverse nesting, and asserts both that the inner type is still checked through the nesting and that required-ness survives the unwrapping (a missing required key still errors). It runs under both the `typing` and `typing_extensions` providers, so it exercises the iterative unwrap rather than the single-qualifier case the old tests covered.

PR #581, opened 2026-08-30
Fix defining class binding in instrumented __new__

### LINKED ISSUE #578: Instrumented `__new__` binds the class's own name to `cls`, so it means the subclass
### Things to check first

### Typeguard version

4.6.0

### Python version

3.14.6

### What happened?

## Summary

class.

## Why it matters

```python
    if cls is Base:
    return super().__new__(cls)
```

## Suggested direction

  gives the wrong class.

`bugdemo/mod.py`:

```python
class Base:
        return super().__new__(cls)


class Sub(Base):
    pass
```

`run.py`:

```python
import sys
if "--hook" in sys.argv:
from bugdemo.mod import Sub
Sub()
```

```console
$ python run.py

$ python run.py --hook
```

### Expected

### Actual

```python
class F:
        return super().__new__(cls)


class SubF(F): pass


SubF()
```

be rejected here.

## Scope

| Case | Result |
|---|---|
| `__init__` | correct |

onuracar-dev:
## Changes

Fixes #578.

- Replace the injected `ClassName = cls` local in instrumented `__new__()` methods with annotation-only resolution through the defining class's `__class__` cell.
- Bind a newly introduced class cell to the actual owner when a whole class is decorated with `@typechecked`; import-hook instrumentation continues to use the compiler-created class cell.
- Add regressions for subclass construction, forward-reference argument checking without a pre-existing class cell, and Enum construction.
- Document the fix in the unreleased changelog.

## Validation

- Regression on upstream: `pytest tests/test_instrumentation.py::test_new_keeps_defining_class_binding -q` failed in both decorator and import-hook modes before the fix.
- Focused final tests: 8 passed.
- Full suite with coverage: 545 passed, 5 skipped, 9 xfailed.
- Pre-commit on all six changed files: passed (Ruff, Ruff format, pinned MyPy, RST, and hygiene hooks).
- Strict documentation build: `python -m sphinx -W -n -E -a ...` passed.
- `uv build`: wheel and sdist built successfully.
- `git diff --check` and staged secret scan: passed.

## Checklist

- [x] You've updated the documentation (in `docs/`, in case of behavior changes or new features)

AI disclosure: OpenAI Codex (GPT-5) analyzed the issue and generated/revised the implementation, tests, changelog text, commit message, and this pull-request description. Separate Codex review passes inspected the full diff and caught a closure-cell defect before publication; the final commit was then revalidated with the commands above. No separate human review is being represented here.


PNHD (changes requested):
One supported decorator path is still broken on this head. SECURITY-N/A.

`owner` is only supplied when `@typechecked` decorates the entire class. When `@typechecked` is applied directly to an individual `__new__` method (which Typeguard documents as a supported function-decorator mode), the transformed annotation now references `__class__`, so the newly compiled code has a `__class__` free variable. But the direct-function path still calls `instrument(target)` with `owner=None`; closure reconstruction therefore falls through to `frame_locals.get("__class__") or ForwardRef("__class__")`, because `__class__` is also recorded in `names_used_in_annotations`.

That means the new function receives a closure cell containing `ForwardRef("__class__")`, not the defining class. I independently reproduced the relevant CPython closure behavior with a reduced compile/FunctionType control: a method compiled with `return __class__` has `co_freevars == ("__class__",)`, and supplying a placeholder cell makes the method load that placeholder. Python's data-model docs likewise specify that `__class__` is a compiler-created closure for methods defined in a class body.

Please add a regression for method-level usage such as:

```python
class Base:
    @typechecked
    def __new__(cls, other: "Base") -> "Base":
        ...
```

including subclass construction / argument checking, and make that path bind the actual defining class rather than `ForwardRef("__class__")`. The existing class-decorator and import-hook regressions can then continue covering the other two instrumentation modes.

Issue #398, opened 2023-09-08
Weird problem with IntEnum


jolaf:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Typeguard version

4.1.3

### Python version

3.10.12

### What happened?

```
Traceback (most recent call last):
  File "test.py", line 4, in <module>
    import_module('A')
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "/home/jolaf/.local/lib/python3.10/site-packages/typeguard/_importhook.py", line 98, in exec_module
    super().exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "A.py", line 2, in <module>
    class A(IntEnum):
  File "/usr/lib/python3.10/enum.py", line 289, in __new__
    enum_member = __new__(enum_class, *args)
  File "A.py", line 3, in __new__
    def __new__(cls, value: int, phrase: str = "") -> "A":
NameError: name 'A' is not defined
```

### How can we reproduce the bug?

`test.py`:
```python3
from importlib import import_module
from typeguard import install_import_hook
with install_import_hook(('A',)):
    import_module('A')
```
`A.py`:
```python3
from enum import IntEnum
class A(IntEnum):
    def __new__(cls, value: int, phrase: str = "") -> "A":
        obj = int.__new__(cls, value)
        obj._value_ = value
        obj.phrase = phrase
        return obj
    v = 5
```

```shell
$ python3 test.py
```

jolaf:
Originally the problem appeared on `import httpx`, I've just narrowed it down.

agronholm:
This happens when `__new__()` is invoked at `v = 5` before the class definition has finished. The interpreter therefore raises `NameError` because the class is not present in the global scope (or anywhere else really) . I don't know if there's anything I can do to fix this.

agronholm:
[...]

PR #583, opened 2026-09-01
Fix __new__ forward reference class aliasing

### LINKED ISSUE #578: Instrumented `__new__` binds the class's own name to `cls`, so it means the subclass

jaideeppyne:
Fixes #578.

When instrumentation sees a string self-reference in `__new__`, it currently rewrites the annotation to the class name and injects `ClassName = cls` into the method body. That avoids the class-creation-time `NameError` from #398, but it leaves a runtime-visible local alias. Later subclass construction then makes `ClassName` resolve to the subclass inside the base `__new__` body.

This changes exact string self-references in `__new__` to `ForwardRef("ClassName")` instead of injecting a local class-name alias. Forward reference evaluation still falls back to `memo.self_type` when the referenced name is not yet resolvable and it matches the class currently being created, preserving the #398 behavior without shadowing user code after class creation.

Added a regression that constructs a subclass under both `@typechecked` and import-hook instrumentation and asserts the base class name still resolves to the base class inside `Base.__new__`.

Validation run locally:

```
uv run --group test pytest -q
uv run ruff format --check src/typeguard/_transformer.py src/typeguard/_utils.py tests/test_transformer.py tests/dummymodule.py tests/test_instrumentation.py
```

`uv run --group test mypy src` is blocked in this local environment by an unrelated missing `typeshed.stdlib.types` import from `src/typeguard/_decorators.py`.


agronholm:
Is this substantially different from #581?

jaideeppyne:
No, not substantially: both address #578 and remove the injected `ClassName = cls` binding. This PR uses a `ForwardRef` for an exact quoted self-reference in `__new__` and a `memo.self_type` fallback in the shared forward-reference evaluator; #581 instead uses the defining class's `__class__` closure and supplies the owner during class decoration.

I missed #581 when opening this. Having compared the patches, #581 also covers subclass construction, argument checking without an existing class cell, and Enum construction, without changing the shared forward-reference evaluator. I don't see a separate use case here that warrants reviewing both implementations, and I defer to #581.

PR #589, opened 2026-09-06
fix: ignore type-only class bindings

### LINKED ISSUE #456: Extend `TYPE_CHECKING` handling to classes and assignments
### Things to check first

### Typeguard version

4.2.1

### Python version

3.12.3

### What happened?

`typechecked` code tries to use a class that is hidden behind `TYPE_CHECKING` and should not be required at runtime.

`bug.py`
```python
# python bug.py && mypy bug.py && pyright bug.py

from __future__ import annotations

from typing import TYPE_CHECKING, TypedDict

# from typeguard import typechecked

if TYPE_CHECKING:

    class Args(TypedDict):
        x: int
        y: int

# @typechecked
def bug() -> None:
    args: Args = {"x": 1, "y": 1}
    print(args)

bug()
```

`python bug.py && mypy bug.py && pyright bug.py`
```
{'x': 1, 'y': 1}
Success: no issues found in 1 source file
0 errors, 0 warnings, 0 informations 
```

Now add the `@typechecked` decorator, and run `python bug.py` again:

```
  File "C:\Code\project\bug.py", line 22, in <module>
    bug()
  File "C:\Code\project\bug.py", line 18, in bug
    args: Args = {"x": 1, "y": 1}
          ^^^^
NameError: name 'Args' is not defined. Did you mean: 'args'?
```

One workaround is to add `bug2.py`:
```python
from typing import TypedDict

class Args(TypedDict):
    x: int
    y: int
```

and modify `bug.py` to
```python
# python bug.py && mypy bug.py && pyright bug.py

from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from bug2 import Args

@typechecked
def bug() -> None:
    args: Args = {"x": 1, "y": 1}
    print(args)

bug()
```

But that feels kind of unnecessary, I think.

be-student:
## Problem

The instrumenter recognized imports inside module-level `TYPE_CHECKING` blocks, but it did not record classes, annotated assignments, or PEP 695 `type` aliases as type-only bindings. With targeted instrumentation from `@typechecked`, unrelated top-level definitions can be removed from the temporary AST before compilation, so annotations that referenced these unavailable bindings could still be wrapped in runtime checks and raise `NameError`.

The same handling also needed to recognize the qualified `typing.TYPE_CHECKING` form while leaving bindings from the runtime `else` branch available for instrumentation.

## Changes

- Collect names bound by classes, annotated assignments, and PEP 695 `type` statements in module-level `TYPE_CHECKING` bodies.
- Collect these bindings before target-specific AST pruning so `@typechecked` does not emit runtime checks for unavailable names.
- Recognize both `if TYPE_CHECKING:` and `if typing.TYPE_CHECKING:` guards.
- Exclude names rebound in the runtime `else` branch from unavailable bindings, including same-name class, annotated-assignment, and type-alias fallbacks. Annotation-only declarations without a value do not count as runtime bindings.
- Add focused regressions for whole-module and targeted instrumentation, simple and generic type aliases, qualified guards, and runtime `else` bindings. Execute transformed modules to verify same-name fallback types reject invalid input.
- Update the feature documentation and unreleased history to describe the supported bindings.

## Validation

Validated at commit `d58c5587f1f2d8a504a1c1bd1da553efb0ec3322`:

- Six same-name runtime fallback regressions failed before the follow-up fix and passed afterward (class, annotated assignment, and PEP 695 alias, each with whole-module and targeted instrumentation).
- Full suite on Python 3.12: 550 passed, 9 skipped, 9 xfailed.
- Full suite on Python 3.10: 522 passed, 37 skipped, 9 xfailed.
- All pre-commit hooks passed, including Ruff formatting/linting and mypy.
- Documentation built successfully with the same two pre-existing unresolved `typing.Union` reference warnings.
- `git diff --check` passed.

## AI disclosure

OpenAI Codex assisted with implementing the maintainer-requested follow-ups, adding regression coverage, and running the validation reported above. The resulting changes were reviewed against the issue and maintainer feedback before this update.

Closes #456


agronholm:
This should be extended to handle the type alias syntax too.

PR #591, opened 2026-09-12
Check Never and NoReturn when nested in another annotation

### LINKED ISSUE #580: Never and NoReturn are not checked when nested inside another annotation, so any value passes
### Things to check first

### Typeguard version

4.6.0 (also present on `main` at `123f1abc7e82bbe980c779005de858533e1786c3`)

### Python version

3.12.14

### What happened?

`Never` and `NoReturn` are handled correctly when either one is the entire argument or return annotation. In the example below `bare(3)` raises `TypeCheckError`, which is what I expected.

When either one appears inside another annotation, no check is performed and every value passes. `list[Never]` accepts `[1]`, `dict[str, Never]` accepts `{"a": 1}`, `Optional[Never]` accepts `1`, and `Sequence[NoReturn]` accepts `[1]`. Nothing is raised and nothing is warned. The two forms disagree with each other in the same argument position:

```
def bare(x: Never)          rejects 3, correct
def nested(x: list[Never])  accepts [1]
```

The case I found most surprising is `Union[int, Never]`, which accepts `"x"`. A plain `str` is not an `int`, so that union passes on its `Never` member alone.

I expected `check_type([1], list[Never])` to raise `TypeCheckError` the same way `check_type(["a"], list[int])` does.

`docs/features.rst` lists `typing.Never` and `typing.NoReturn` as "Supported in argument and return type annotations", and `list[Never]` is an argument annotation, so I could not work out from the docs which of the two behaviours is intended. If nesting is deliberately out of scope, then narrowing that table entry would have saved me the confusion, because the bare form working is exactly what led me to expect the nested form to work.

This is separate from #579. That one is about PEP 695 aliases; this reproduces with no alias involved, and the bare form here already works.

Standard library only, no third party packages beyond typeguard itself.

```python
from typing import Never, NoReturn, Optional, Sequence, Union

from typeguard import TypeCheckError, check_type, typechecked


@typechecked
def bare(x: Never) -> int:
    return 1


@typechecked
def nested(x: list[Never]) -> int:
    return 1


try:
    bare(3)
    print("bare(3)        accepted")
except TypeCheckError:
    print("bare(3)        TypeCheckError")

try:
    nested([1])
    print("nested([1])    accepted")
except TypeCheckError:
    print("nested([1])    TypeCheckError")

for annotation, value in [
    (list[Never], [1]),
    (dict[str, Never], {"a": 1}),
    (Union[int, Never], "x"),
    (Optional[Never], 1),
    (tuple[Never, ...], (1, 2)),
    (Sequence[NoReturn], [1]),
]:
    try:
        check_type(value, annotation)
        print(f"check_type     accepted  {value!r:12} against {annotation}")
    except TypeCheckError:
        print(f"check_type     TypeCheckError  {value!r} against {annotation}")
```

Output on 4.6.0 under Python 3.12.14, in a container with only typeguard installed:

```
bare(3)        TypeCheckError
nested([1])    accepted
check_type     accepted  [1]          against list[typing.Never]
check_type     accepted  {'a': 1}     against dict[str, typing.Never]
check_type     accepted  'x'          against typing.Union[int, typing.Never]
check_type     accepted  1            against typing.Optional[typing.Never]
check_type     accepted  (1, 2)       against tuple[typing.Never, ...]
check_type     accepted  [1]          against typing.Sequence[typing.NoReturn]
```

The first line is the control and is correct. Every line after it is a value passing a check I expected it to fail. The output is byte for byte the same at `main` HEAD `123f1abc`.

dchaudhari7177:
## Changes

Fixes #580.

`Never`/`NoReturn` were recognised only as an *entire* argument or return annotation, in `_functions.py`. Nested inside another annotation, `builtin_checker_lookup` found no checker, so `check_type_internal` fell off the end of its dispatch and returned — silently, with nothing warned. `Never` is not a class and not a string, so neither of the two fallbacks after the lookup applied either.

`_checkers.py` now registers a checker for both (`typing.Never` on 3.11+, `typing.NoReturn`, and the `typing_extensions` re-exports, following the existing comment about not assuming which version re-exports what).

Before / after on the reproduction from the issue:

| annotation | value | before | after |
| --- | --- | --- | --- |
| `list[Never]` | `[1]` | accepted | `item 0 of list is not compatible with Never` |
| `dict[str, Never]` | `{"a": 1}` | accepted | `value of key 'a' of dict is not compatible with Never` |
| `Union[int, Never]` | `"x"` | accepted | fails both union members |
| `Optional[Never]` | `1` | accepted | fails both union members |
| `tuple[Never, ...]` | `(1, 2)` | accepted | `item 0 of tuple is not compatible with Never` |
| `Sequence[NoReturn]` | `[1]` | accepted | `item 0 of list is not compatible with Never` |

A `Never` that no value reaches is still satisfied rather than violated — `list[Never]` accepts `[]`, `Union[Never, None]` accepts `None`, `Union[int, Never]` accepts `3` — because the check runs per value, through the existing container checkers, rather than on the container itself. That felt like the right reading: an empty `list[Never]` is inhabited, and rejecting it would make the type unusable as a "this collection is always empty" annotation.

The message names the type rather than the call. Nested there is no call to talk about, so `_functions.py`'s "declared never to be called" / "declared never to return" wording does not fit; the bare form keeps it and is unchanged.

On the docs point in the issue: the `features.rst` row said "Supported in argument and return type annotations", which is what led you to expect nesting to work. Since nesting now does work, the row says what happens instead.

## Checklist

- [x] You've updated the documentation (in `docs/`, in case of behavior changes or new features)

### Verification

`TestNever` in `tests/test_checkers.py` — 10 cases: the four nested containers, the `Union` member, the bare form, and four that must keep passing. With `src/typeguard/_checkers.py` reverted, 6 of the 10 fail with `DID NOT RAISE TypeCheckError`, and the four "nothing to check" cases pass either way, which is what makes them worth having.

Full suite: **541 passed, 9 skipped, 9 xfailed.** The 4 failures in `tests/test_pytest_plugin.py` are pre-existing on an unmodified checkout (`pytest-asyncio`'s unset `asyncio_default_fixture_loop_scope` warning raised as an error) and unrelated.

`ruff` 0.15.20 (the pinned version) check and format clean; `mypy` clean on the changed module.


dchaudhari7177:
Closing as a duplicate of #582, which predates this by twelve days and fixes the same issue. I should have checked the open PRs before writing this — sorry for the extra review load.

If anything here is useful to fold into #582, it is the four cases asserting what must keep passing: \`list[Never]\` with \`[]\`, \`dict[str, Never]\` with \`{}\`, \`Union[Never, None]\` with \`None\`, and \`Union[int, Never]\` with \`3\`. They pass both before and after the fix, so they are not regression tests for the bug — they pin the boundary, i.e. that the check runs per value rather than rejecting the container outright.

PR #593, opened 2026-09-15
Check annotated locals bound by for and with targets

### LINKED ISSUE #586: Check annotated locals bound by `for` and `with` targets
Reframed as a feature request, per agronholm's comment below. Originally filed as a bug report.

Typeguard checks the value assigned to an annotated local by `=`, by the walrus operator, by chained assignment and by tuple unpacking. It does not check a value bound to the same annotated name by a `for` target or a `with ... as` target. Same for `async for`, `async with ... as`, `except ... as`, `except* ... as`, a `match` capture pattern and `import ... as`.

The request is to extend the local-variable check to those binding forms, or to whichever subset is worth the transformer complexity. `for` and `with` are the two that turn up in ordinary code.

Same behaviour on 4.6.0 and on main at `123f1abc7e82bbe980c779005de858533e1786c3`, python 3.11.16 / 3.12.14 / 3.13.15. `install_import_hook` behaves like `@typechecked` here, so it is not specific to the decorator. mypy flags all three cases below as `Incompatible types in assignment`.

```python
import io

from typeguard import TypeCheckError, typechecked


@typechecked
def plain_assign() -> None:
    x: int
    x = "not an int"


@typechecked
def for_target() -> None:
    x: int
    for x in ["not an int"]:
        pass


@typechecked
def with_target() -> None:
    x: io.BytesIO
    with io.StringIO() as x:
        pass


for func in (plain_assign, for_target, with_target):
    try:
        func()
        print(f"{func.__name__}: no error")
    except TypeCheckError as exc:
        print(f"{func.__name__}: TypeCheckError: {exc}")
```

```
plain_assign: TypeCheckError: value assigned to x (str) is not an instance of int
for_target: no error
with_target: no error
```

Ananthr16:
Fixes #586.

## Problem

Typeguard checks the value assigned to an annotated local by `=`, the walrus operator, chained assignment, and tuple unpacking -- but not a value bound to the same annotated name by a `for` target or a `with ... as` target (or their async variants). So this silently skips type checking:

```python
@typechecked
def process(values: list[int]) -> None:
    item: int
    for item in values:  # not checked -- a str, a dict, anything passes silently
        ...
```

Originally filed as a bug report; the maintainer reframed it as a feature request ("Not a bug... I can take this into the development queue if it's framed as a feature request").

## What changed

`src/typeguard/_transformer.py`: added `visit_For`/`visit_AsyncFor` and `visit_With`/`visit_AsyncWith` to `TypeguardTransformer`, plus a shared `_binding_checks` helper that recurses into `Tuple`/`List`/`Starred` targets (mirroring how `visit_Assign` already handles tuple-unpacking) and looks up each bound `Name` in the existing `self._memo.variable_annotations` table built up by `visit_AnnAssign`.

Two things needed care beyond just "insert a check call":

1. **Timing of the annotation lookup.** For `for` loops, `_binding_checks` is called on `node.target` *before* `self.generic_visit(node)` walks the loop body -- so only annotations already known before the loop is entered apply. An annotation written *inside* the loop body must not retroactively apply to the loop target bound before it (covered by `test_for_loop_later_annotation_not_retroactive`).

2. **`with` items aren't independent.** A `with a, b:` statement is equivalent to nested `with a: with b:` -- `b`'s context manager is only entered after `a`'s `__enter__` (and, now, `a`'s binding check) completes. If `a`'s check fails, `b` must never be entered at all. Since there's no single expression to wrap for a `with` target (the binding happens via the interpreter's own `__enter__`/`__aenter__` call, not an expression we control), `visit_With` rebuilds the statement into a chain of single-item `with` blocks, one per item, each wrapping the next -- so each binding's check runs (and can raise) before the next item's context manager is entered. This is exercised by `test_with_binding_first_target_fails_before_second_manager_entered`, which records enter/exit order across two context managers and asserts the second is never entered when the first's check fails.

Also covered: non-`Name` targets (e.g. `for obj.attr in values:`) are left uninstrumented rather than crashing the transformer (there's nothing to look up in `variable_annotations` for a non-`Name` target), and module-level `for`/`with` bindings are left uninstrumented, matching the existing behavior for plain assignments (only function-local bindings are checked).

`docs/features.rst` and `docs/versionhistory.rst` updated accordingly.

### Explicitly out of scope

Per the issue's own framing, I didn't attempt exception handlers (`except ... as`), `match`/`case` capture patterns, `import ... as`, or comprehension targets -- these are called out in `features.rst` as still-unsupported binding forms. Each has its own instrumentation-timing questions (e.g. an `except ... as` name is unbound again at the end of the handler) that felt worth keeping separate from this PR.

## Prior attempt

#588 covered the same ground and is a good reference for the general shape of this fix, but it was closed by its own author with no maintainer review. I read through its diff before writing this, and independently wrote my own implementation and test suite rather than reopening or reusing it -- the approach (checking bound names right after native binding, via new `NodeTransformer` visitor methods) ends up similar, which isn't a coincidence given the constraints of the problem, but the code itself is my own.

## How it was verified

- `pytest`: **563 passed, 9 skipped, 9 xfailed** (up from 535/9/9 on unmodified `master` -- the +28 is exactly the new parametrized tests across the `typechecked`/`importhook` instrumentation methods; zero regressions).
- `coverage run -m pytest` + `coverage report --include='*/_transformer.py'`: **97% (605 stmts, 20 missing)**, and the 20 missing lines are the identical pre-existing gaps present on unmodified `master` (same count, same relative locations) -- every line I added is covered, including the non-`Name`-target and module-scope-noop branches (covered by two dedicated tests).
- `ruff check` / `ruff format --check` on the changed files: clean.
- `mypy src/typeguard`: one pre-existing, unrelated error (`typeshed.stdlib.types` stub not found in this environment) that reproduces identically on unmodified `master` -- not something introduced by this change.
- `sphinx-build -b html docs build/sphinx/html -W` (warnings-as-errors): builds cleanly with the updated `features.rst`/`versionhistory.rst`.


docs/features.rst
```diff
@@ -12,13 +12,19 @@ The following type checks are implemented in Typeguard:
...
+  binds the target -- before the loop body runs, or before the next context
+  manager in the same ``with`` statement is entered.
 
 What does Typeguard NOT check?
 ------------------------------
 
 The following type checks are not yet supported in Typeguard:
 
+* Local variable bindings in exception handlers (``except ... as``), pattern
+  matching (``match``/``case``), imports (``import ... as``) and comprehensions
```
agronholm: Updating the negatives is not really in the scope of this PR.

```suggestion
```

docs/features.rst
```diff
@@ -12,13 +12,19 @@ The following type checks are implemented in Typeguard:
 * Types of values returned from instrumented functions
 * Types of values yielded from instrumented generator functions
 * Types of values sent to instrumented generator functions
-* Types of values assigned to local variables within instrumented functions
+* Types of values assigned to local variables within instrumented functions,
+  including previously annotated names bound by ``for``, ``async for``,
+  ``with ... as`` and ``async with ... as``. These are checked right after Python
+  binds the target -- before the loop body runs, or before the next context
+  manager in the same ``with`` statement is entered.
```
agronholm: Separate these to their appropriate bullet points and drop the prose.

